PR #27, opened 2026-02-15
Group15: Core c++ classes

GumdropHorse:
Creation of Group 15's core c++ classes and their test files: DataFileManager, Random, WeightedSet, WorldGrid and World Position.


source/tools/Random.hpp
```diff
@@ -0,0 +1,245 @@
...
+                    return false;
+                }
+                return true;
+            }
+
+            // Generates based off of a given probability, and returns a bool
+            bool P(double probability = 0.5){
+                // Error handling: probability must be between 0 and 1
+                if (0 > probability || 1 < probability) {
+                    throw std::runtime_error("cse498::Random::GetP(): probability must be between 0 and 1.");
```
Lak453: In your P() function, the exception message says GetP() but your function is named P(). This is minor, but it’s worth fixing because error messages should match the API name to help debugging.

source/tools/WeightedSet.hpp
```diff
@@ -0,0 +1,152 @@
...
+		void Insert(const T& id, double weight){
+			if (weight < 0.0) {
+				throw std::invalid_argument("cse498::WeightedSet::Insert(): weight must be non-negative");
+			}
+			if (item_idx.contains(id)) {
+				throw std::invalid_argument("cse498::WeightedSet::Insert(): duplicate item");
+			}
+
+			//Tiny weights below tolerance that are not 0 can cause strange sample outputs
+			if (weight < TOL) {
```
hoangtu0701: This tolerance-clamping block is the same in both Insert and Update. Could pull it out into a small helper to keep things DRY

source/tools/WeightedSet.hpp
```diff
@@ -0,0 +1,152 @@
...
+					break; //return this node since there is no left or right
+				}
+				outer_lo = inner_up;
+				idx = r_idx;
+			}
+			return items[idx];
+		}
+
+		int GetSize() const { return size; }
+		double GetItemSum(T item) { return sum_tree[item_idx[item]]; }
```
hoangtu0701: These getters (GetItemSum and GetWeight) should probably be const. Also taking T by const ref instead of by value would avoid extra copies if T is something like std::string.

source/tools/Random.hpp
```diff
@@ -0,0 +1,245 @@
...
+            // Ensursed state values are non-zero and well-mixed.
+            uint64_t splitmix64(struct splitmix64_state *state) {
+                uint64_t result = (state->s += GOLDEN_RATIO);
+                result = (result ^ (result >> RIGHT_SHIFT1)) * FIRST_MIXING;
+                result = (result ^ (result >> RIGHT_SHIFT2)) * SECOND_MIXING;
+                return result ^ (result >> RIGHT_SHIFT3);
+            }
+
+            // Uses m_seed to generate the state positions
+            void xoshiro356pp_init(struct xoshiro256pp_state *state) {
```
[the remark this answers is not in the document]
GumdropHorse: Thank you for the feedback! This is not a typo. _init is the function called to start generating the rng, _state is the created rng based off of the seed, and xoshiro256pp actually generates a number. xoshiro256pp is the algorithm that I based this off of and credited.

source/tools/Random.hpp
```diff
@@ -0,0 +1,245 @@
...
+                m_seed =  std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+            }
+            virtual ~Random() = default;
+
+            // Public getter and setter for the current seed.
+            void SetSeed(uint64_t seed) {
+                m_seed = seed;
+                m_used = false;
+            }
+            uint64_t GetSeed() {return m_seed;}
```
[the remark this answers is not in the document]
GumdropHorse: Thank you for catching this, resolved!

source/tools/Random.hpp
GumdropHorse: Thank you for catching this, resolved!

source/tools/WeightedSet.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+	template <typename T>
+	class WeightedSet {
+		private:
+		//Storage should be relatively small assuming T will be a string or int (probable for ids)
+		std::unordered_map<T, int> item_idx{};
+		std::vector<T> items{};
+		std::vector<double> weights{};
+		std::vector<double> sum_tree{}; 
+		int size = 0;
```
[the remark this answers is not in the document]
AbbyFranz: Fixed!

source/tools/WeightedSet.hpp
AbbyFranz: Fixed!

source/tools/Random.hpp
Mehver: # 1 Class Design

## 1.1 Informative identifiers

Are names of variables, functions, types, etc, clear and helpful for a reader to understand the code. For example, if you need to track the number of students in a class, good variable names might be `num_students` or `student_count`, whereas you don’t want something too short to be confusing (like `N` or `num_s`) or too verbose that it’s annoying (like `total_number_of_students_in_class`).

> [!IMPORTANT]
>
> Everything looks good; it’s just that the `xoshiro356pp_init` here might have been intended to be 256 rather than 356?
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L70-L78

## 1.2 Avoids “magic numbers”

Similar to above, you should use constant variables with informative names instead of arbitrary numbers throughout your code, and instead set constants. For example, I have a virtual CPU class with 6 virtual registers, but rather than using “6” throughout, I set `static constexpr size_t NUM_REGS = 6;` at the top of the class to make it clearer what the 6 refers to (and, in this case so that I can change it easily if needed.) Note that if the meaning of the numbers is obvious from context, it may not be an issue, such as initializing a reference count to 1 in a constructor, or an accumulator to zero.

>[!IMPORTANT]
>
>These `constexpr` usages are appropriate, but you might consider placing them inside the `namespace cse498` as well. And if possible, use `static constexpr`.
>
>https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L17-L35

## 1.3 Intuitive and functional interface

Does the class expose only the necessary functionality, and is it easy to use correctly while hard to misuse? Are member variables private? Is there an obvious function to use for common tasks, with relevant default arguments for parameters? Are any of the functions misleading, like potentially confusing uses of operator overloading? (Feel free to provide suggestions on how to improve the interface for better usability)

> [!NOTE]
>
> Everything else looks good. As for default parameters, my personal view is that they can sometimes encourage caller-side laziness, causing call sites that should explicitly specify arguments to omit them. This can lead to unexpected issues during future migrations or refactoring. Please use your own judgment to decide whether they should be removed.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L162
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L180
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L198

> [!IMPORTANT]
>
> Unless you explicitly intend the class to be inherited from and deleted via a base-class pointer, a virtual destructor implies that this is a base class meant for polymorphic use. If that is not the design goal, consider removing `virtual`, or declare the class as non-inheritable (eg `final`) to better express the intent.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L132C12-L132C41

## 1.4 Informative comments

Do comments help someone new to the code understand it more rapidly? Did the coder properly avoid redundant or obvious comments that clutter the code (e.g., `++x; // increments x`). In general, comments should focus on *why* code is helpful, rather than belabor *what* that code does (though with complex code, the latter is useful too.)

> [!TIP]
>
> The comments are all good. If I had to make a suggestion, as these are utility functions, it might be worth discussing with the API Officer or Documentation Officer whether a unified IDE-friendly comment format could be established, so that callers can quickly see the specific requirements for the function parameters.

> [!NOTE]
>
> You can briefly explain the purpose of each `#include` with comments.

> [!IMPORTANT]
>
> For cases like this, it’s best to explain the valid range of the parameters in the comments.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L227-L232

## 1.5 Consistent style

Are style choices (like naming schemes, brace placement, etc) used in a consistent and informative manner throughout the code? Is there consistency in how different naming conventions are used? (e.g., `camelCase`, `snake_case`, `ALL_CAPS`, or `PascalCase`, or prefixes and suffixes, like `m_` at the beginning of a member variable.).

> [!TIP]
>
> Very good.

## 1.6 Visual appearance

Is the code easy to scan through and visually well structured with good use of whitespace, indentation, and line breaks? Do there appear to be consistent limits on line widths?

> [!IMPORTANT]
>
> The spacing around the `*` in the pointer here could be made consistent.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L63
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L71
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L86

> [!IMPORTANT]
>
> Also the blank lines here seem a bit excessive.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L42-L45

# 2 Implementation

## 2.1 No warnings during compilation

Does the code compile with the `-Wall` and `-Wextra` flags without producing warnings? One option is to add `-Werror` to your list of options, which turns all warnings into errors and thus halts compilation when one occurs.

> [!TIP]
>
> Very good.

## 2.2 Proper uses of namespaces

Is all code within `namespace cse498 { ... }` ? Does the code avoid the generic `using namespace std` as well as avoiding any namespace pollution in header files that will spread to users of those files.

> [!IMPORTANT]
>
> Ref to the issue mentioned in 1.2.

## 2.3 Adheres to the C++ standard

Does the code avoid undefined behavior? If possible, make sure the code compiles and runs properly on different compilers and different platforms.

> [!IMPORTANT]
>
> If `k == 0` or `k == 64`, it will trigger shift undefined behavior (in C++, shifting by a count equal to or greater than the type width is UB). The current call sites use the constants `LEFT_SHIFT = 23` and `NUM_STATE_VALUES = 45`, so it is safe as is, but the interface itself does not guard against misuse.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L80-L83

## 2.4 Appropriate Modularity

Are common procedures grouped into useful helper functions? Are helper classes used where needed? Is unnecessary code duplication avoided?

> [!IMPORTANT]
>
> This section could be grouped into a helper function.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L148-L150
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L168-L171
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L186-L189
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L204-L207
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L234-L237

## 2.5 Uses modern C++ constructs

Where appropriate, does the code use range-based for loops, `std::optional`, `std::expected`, or other modern C++ techniques? (In the future we will also look for proper use of templates, move semantics, lambdas, and `constexpr`, but only once we cover those topics).

> [!IMPORTANT]
>
> It is recommended to replace `<stdint.h>` with `<cstdint>`, as the latter offers better compatibility and is more strongly recommended. Also, `<cstdbool>` is the C++ counterpart of `<stdbool.h>`, but in C++ `bool`/`true`/`false` are built-in, so including it is usually unnecessary.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L12
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L14

## 2.6 Uses standard library algorithms

Does the code avoid extraneous loops when equivalent functionality exists in the standard library? For example, `std::fill` will fill a vector, `std::remove_if` will remove elements from a container based on a condition, or `std::is_sorted` will tell you if the values in a container are in order. See https://en.cppreference.com/w/cpp/algorithm for a full list (it’s LONG, but well worth being familiar with).

> [!TIP]
>
> Very good.

## 2.7 Pointer handling

Is memory management done carefully, with either smart pointers (`std::unique_ptr`, `std::shared_ptr`) or, only if needed, careful use of raw pointers with `new` and `delete`?

> [!NOTE]
>
> If it doesn’t affect the functionality, you may consider changing the pointer to a reference to reduce the risk of `nullptr`.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L63
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L71
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/df3cfe6886b051144e939c6a8245a864490c61a7/source/tools/Random.hpp#L86

## 2.8 `const` correctness

Are member functions `const` if they don’t change the state of the class? Are parameters (especially reference parameters) marked `const` if they don’t change inside the function?

> [!TIP]
>
> Very good.

# 3 Testing and Error Checking

## 3.1 Unit testing

Are unit tests present for all public member functions (including constructors) and well organized? Do they trigger cases that will exercise most private portions of the class’ code?

> [!IMPORTANT]
>
> 1. Although `Random` doesn’t manage explicit resources, you can still test the behavior of copy/assignment/move operations.
> 2. In addition, you can test reproducibility with the same seed.
> 3. Also, each `TEST_CASE` currently packs many different functions and scenarios, which makes failures harder to localize. Consider splitting them into multiple `TEST_CASE`s by function or using `SECTION()` to organize them.
> 4. And, consider adding `#include <algorithm>` in the test since you are using `std::any_of`.

## 3.2 Edge cases

Does testing cover most edge cases, as possible? For example, for pointers, does it test null pointers (if they are allowed)? Does it test for empty containers being used? Does it test for 0 or other boundary conditions on parameters?

>[!IMPORTANT]
>
>It is recommended to provide more potentially ill-formed parameter cases that may occur in real usage to ensure the parameter validation is sufficiently robust.

## 3.3 Assertions

Are `assert` statements used properly throughout the code to enforce invariants and ensure a valid state? (for example, checking that valid arguments are passed into functions.) Note that asserts should be used only to identify conditions that should never occur in properly working code; if code is expected to recover, error handling must be performed.

>[!TIP]
>
>Very good.

## 3.4 Error handling

Are error conditions that can occur during correct execution (such as resource limitations, files not available, or division by zero) checked for and reported back with `std::expected`, other return conditions, or some other technique like exceptions (if your group is comfortable with the speed hit)?

>[!TIP]
>
>Very good.

Mehver (commented):
I’ve put all the comments into a single document for your convenience.

source/tools/WorldGrid.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+#include <cassert>
+#include <fstream>
+#include <iostream>
+#include <unordered_map>
+#include <vector>
+
+#include "../core/Location.hpp"
+#include "../tools/io_utils.hpp"
+#include "../tools/io_utils.hpp"
```
motowsk2: Consider getting rid of unused headers like <fstream> and <iostream> if they aren't needed. This helps decrease compile time and doesn't pollute global namspace. Also, there is a duplicate io_utils.hpp include that can be removed.

source/tools/WorldGrid.hpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    // Build a map that will convert loaded char symbol into their cell type ID.
+    auto BuildSymbolMap() {
+      std::unordered_map<char, size_t> symbol_map;
+      for (size_t i=0; i < cell_types.size(); ++i) {
+        symbol_map[cell_types[i].symbol] = i;
+      }
+      return symbol_map;
+    }
```
motowsk2: This function should probably be const and not public. Correct me if you think I am wrong, but it looks like it doesn't modify the object and is only used internally by Load(). Could just be updated to something like this [[nodiscard]] auto BuildSymbol Map() const

tests/tools/GenerativeWorld/WorldGrid.cpp
motowsk2: # 1. Unit Testing

The existing tests cover basic construction, dimension accessors, default cell values, and `IsValid(double, double)`, which is a solid start. However, many public methods (`AddCellType`, `GetCellTypeID`, `Resize`, `GetSymbol`, `WorldPosition` overloads of `IsValid`, and the `Load`/`Print` functions) are not included in the test suite.

To fully meet the rubric, additional unit tests should be added to cover these parts of the public API.

# 2. Edge Cases

The `IsValid` tests cover in-bounds corners and some out-of-bounds cases, which is very helpful. You could expand these edge cases by testing empty grids, 1×1 grids, resizing behavior (both shrinking and expanding), unknown cell type symbols and/or names, and testing uneven/different-length input rows.

Adding these would further satisfy the rubric’s emphasis on edge cases.

# 3. Assertions

Assertions should be added to `ToIndex` and the `WorldPosition` overloads, as mentioned in a previous comment.

# 4. Error Handling

The current tests verify that the unknown type exists but do not test potential error scenarios. These include lookups of unknown names or out-of-range IDs. Adding tests that validate these behaviors—such as `GetCellTypeID("nonexistent") == 0` or verifying that invalid IDs return the unknown type—would improve coverage of the error-handling strategy described by the rubric.

Mehver (commented):
For Random Test's review please also ref to this comment: https://github.com/CSE498/Spring2026-CompanyA/pull/27#discussion_r2839076229

source/tools/Random.hpp
GumdropHorse: Thank you so much for the thorough review! I just went through and all of your suggested edits/changes are now resolved!

motowsk2 (commented):
Submitting my comments for review, overall great job! Only minor things in your WorldGrid.hpp file, would focus more on adding to your test sutie.

source/tools/WorldPosition.hpp
CShon: Overall review:

Class Design

- Informative identifiers
Class, member, and method names are clear, concise, and intuitive.

- Avoids "magic numbers"
Generally speaking uses of numeric values are probably okay given that the context looks clear, but to note:
For the `Up()`, `Down()`, and similar functions, a small constant such as `double CELL_STEP = 1.0;` could improve clarity.
This is just a nitpick and I don't think it's an issue, but do ask course instructors for their opinion.

- Intuitive and functional interface
Member variables are private, constructors are flexible, accessors are clearly named.
Direction access via `GetDir()`/`SetDir()` and position access via `X()`/`Y()` and `CellX()`/`CellY()` are safe.

- Informative comments
As someone reviewing this code without knowing the context in how it will be used, the comments are generally informative and aren't inherently obvious.
However, comments explaining things like the difference between `X()`/`Y()` and `CellX()`/`CellY()` and how the moving a coordinate position with `Up()`/`Down()`/`Left()`/`Right()` works relative to the screen coordinates would be helpful.

- Consistent style
Brace placement, naming style, and comments formatting is consistent throughout the file, except for `isColliding()` which differs from `IsAdjacentCell()` in the capitalization.

- Visual appearance
I personally had no problems when scanning over your code and identifying the structural components of your implementation. No issues here.

Implementation

- No warnings during compilation
No issues here.

- Proper uses of namespaces
No issues here.

- Adheres to the C++ standard
No issues here.

- Appropriate modularity
Helper methods `CellX()`/`CellY()` are reused properly.

- Uses modern C++ constructs
No issues here.

- Uses standard library algorithms
Algorithms are outside the problem domain, so no issues here.

- Pointer handling
No pointers are in this implementation.

- const correctness
No issues here.

source/tools/WorldPosition.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    [[nodiscard]] double X() const { return x; }
+    [[nodiscard]] double Y() const { return y; }
+
+    [[nodiscard]] Orientation GetDir() const {return dir;} // Returns the current facing direction
+    void SetDir(Orientation new_dir) {dir = new_dir;} // Sets a new facing direction
+
+    [[nodiscard]] size_t CellX() const {
+      assert(x >= 0.0);
+      return static_cast<size_t>(x);
```
alexelkins23: For these casts you could consider using the `gsl::narrow_cast` we learned about in class to be explicit that we are intentionally losing precision.

source/tools/WorldPosition.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+
+    // DEVELOPER NOTE: Add a SameCell function to identify if two positions are in the same cell.
+    [[nodiscard]] bool isColliding(const WorldPosition& other) const
+    {
+      return (CellX() == other.CellX() && CellY() == other.CellY());
+    }
+
+    ///Check if two positions are in adjacent cells
+    [[nodiscard]] bool IsAdjacentCell(const WorldPosition& other) const
```
alexelkins23: It looks like this function repeats a lot of the work from `CellDistance()`. I think this could be simplified to `return CellDistance(other) == 1`.

source/tools/WorldPosition.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    [[nodiscard]] Orientation GetDir() const {return dir;} // Returns the current facing direction
+    void SetDir(Orientation new_dir) {dir = new_dir;} // Sets a new facing direction
+
+    [[nodiscard]] size_t CellX() const {
+      assert(x >= 0.0);
+      return static_cast<size_t>(x);
+    }
+    [[nodiscard]] size_t CellY() const {
+      assert(y >= 0.0);
```
alexelkins23: In `CellX()` and `CellY()` it looks like you don't wan't to allow negative cells, but there is nothing preventing negative positions. If you want to allow negative positions, I would recommend using `std::expected` here to indicate the position is not in a cell, or asserts in the constructor and other functions to not allow negative positions.

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+namespace cse498 {
+
+  class DataFileManager
+  {
+  private:
+    std::string m_filename;
+
+    //MazeWorld *m_world; // This will be a unique_ptr stored in the world file once that file is complete
+    WorldBase *m_world; // This will be a unique_ptr stored in the world file once that file is complete
```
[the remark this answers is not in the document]
lunncurt: Agreed, see line 127 for an example of where a null check could/should be.

lunncurt (commented):
DataFileManager is well implemented and constructed, but I would recommend extra 'worst case' testing / error checking, null ptr checks, and Update() return state (see comments on DataFileManager.cpp/hpp).

source/tools/WorldPosition.hpp
```diff
@@ -0,0 +1,94 @@
...
+      return static_cast<size_t>(y);
+    }
+
+    /// Enable all comparison operators (==, !=, <, <=, >, >=)
+    auto operator<=>(const WorldPosition &) const = default;
+
+
+    // DEVELOPER NOTE: Add a SameCell function to identify if two positions are in the same cell.
+    [[nodiscard]] bool isColliding(const WorldPosition& other) const
```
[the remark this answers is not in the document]
joeyhyun1: Done, thank you

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,142 @@
...
+      WorldGrid &grid = m_world->GetGrid();
+      auto cells = grid.BuildSymbolMap();
+
+      std::string tiles = StoreData(0, "Tile", cells);
+      PacingAgent agent(0, "Pacer", *m_world);
+      std::string agents = StoreData(1, "Agent", m_world->GetKnownAgents(agent));
+
+      file << tiles << "\n" << agents << "\n";
+      file.close();
+    }
```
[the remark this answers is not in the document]
ZhixiangMiao: Fixed it.

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+        else {
+            throw std::runtime_error("cse498::DataFileManager::data_to_string(): Unconvertible data type");
+        }
+    }
+
+    
+
+  public:
+    // Constructor to initialize the DataFileManager with a filename and a reference to the world. 
+    DataFileManager(const std::string & filename, WorldBase *world) :
```
nahumj: This is a raw pointer pointer, not a reference to WorldBase

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+        data_stored += data_string;
+
+        return data_stored;
+    }
+
+    // Function to update the data file with the current state of the world, including tile and agent information. 
+    bool Update()
+    {
+      std::ofstream file;
+      file.open(m_filename, std::ofstream::app);
```
nahumj: Why are you opening the file in append mode?

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+      std::string agents = FormatData(AGENT_ID, AGENT_TYPE, m_world.GetKnownAgents(agent));
+
+      file << tiles << "\n" << agents << "\n";
+
+      file.close();
+
+      return true;
+    }
+
+  protected:
```
nahumj: Why is this here?

source/tools/DataFileManager.hpp
nahumj: What is the point of this class? Is seems to only be able to write to a file some data from a specific world class. It doesn't have any ability to read data from a file (and I suspect it would be difficult to do so because of the novel file format).

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+// For demonstration purposes, we are saving the data from MazeWorld
+#include "../../source/Worlds/MazeWorld.hpp"
+#include "../../source/core/WorldBase.hpp"
+#include "../../source/core/WorldGrid.hpp"
+#include "../../source/Agents/PacingAgent.hpp"
+
+
+namespace cse498 {
+  constexpr int TILE_ID = 0;
```
nahumj: Why do you have all these global variables?

source/tools/Random.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+            ////////////////////////////////////////////////////////////
+            ///
+            /// Xoshiro generation functions
+            /// adapted from the credited source above
+            ///
+
+            // keeps an internal state of 4 64 unsigned 64-bit ints
+            // changes each time a number is generated
+            struct xoshiro256pp_state {
```
nahumj: Inconsistent style of member function names.

tests/tools/GenerativeWorld/Random.cpp
```diff
@@ -0,0 +1,720 @@
...
+    CHECK(std::any_of(generations_p_large.begin() + 1, generations_p_large.end(), [&](const auto& x) { return x != generations_p_large[0]; }));
+
+    // With small error range (ill-formed parameter test cases)
+    CHECK_THROWS(ran14.P(-1));
+
+    // With large error range (ill-formed parameter test cases)
+    CHECK_THROWS(ran14.P(1.1));
+}
+
+TEST_CASE("Test Random Integer Generation Reproducability with Same Seed", "[core]") {
```
nahumj: This is a useful test case, almost all of the previous test cases have little value as they are just testing if the RNG sometimes returns a different value after 1000 draws.

nahumj (commented):
Please see your private channel on discord for your grade,

nahumj (approved):
Please see your private channel on discord for your grade,

source/tools/DataFileManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    // Getter for the filename, allowing retrieval of the current filename being used by the DataFileManager. 
+    std::string GetFilename() const { return m_filename; }
+
+    template <typename T> 
+
+    // Function to store data in a specific format, taking an ID, a type, and the data itself. 
+    std::string FormatData(int id, std::string_view type, const T & data) const {
```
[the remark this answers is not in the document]
ZhixiangMiao: Moved it under private, the id was unnecessary and removed. This function tries to add a tab after writing a block of data into the file, the tab separates between the previous block of data and the next block of data.

source/tools/Random.hpp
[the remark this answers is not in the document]
GumdropHorse: Thank you, resolved!

source/tools/Random.hpp
```diff
@@ -0,0 +1,243 @@
...
+ * It is apart of Green and White Games (Company A) Group 15's module.
+ * @brief A class for random generation
+ * @note Status: PROPOSAL
+ * Credit notes: This files makes use of the Xoshiro algorithm as described at:
+ *               https://en.wikipedia.org/wiki/Xorshift#xoshiro
+ **/
+
+#pragma once
+
+#include <stdint.h>  // This defined the size of uint64_t
```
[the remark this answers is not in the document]
GumdropHorse: Thank you, resolved!

source/tools/Random.hpp
```diff
@@ -0,0 +1,243 @@
...
+#include <chrono>    // Used to get the current time off of the computer
+#include <stdexcept> // Allows the program to throw runtime errors
+
+
+namespace cse498 {
+
+    // Constants
+    constexpr int STATE_NUMBER = 4;
```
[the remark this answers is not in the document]
GumdropHorse: Thank you, resolved!

source/tools/Random.hpp
GumdropHorse: Thank you, resolved!

source/tools/Random.hpp
```diff
@@ -0,0 +1,243 @@
...
+            bool m_used = false;
+
+            void check_rng() {
+                if (m_used == false) {
+                    xoshiro256pp_init(m_rng);
+                    m_used = true;
+                }
+            }
+
+        protected:
```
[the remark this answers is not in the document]
GumdropHorse: This has been deleted now.
Thank you, resolved!

tests/tools/GenerativeWorld/Random.cpp
```diff
@@ -0,0 +1,720 @@
...
+ **/
+
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "../../../source/tools/Random.hpp"
+
+#include <vector>
+#include <algorithm>
+
+
+TEST_CASE("Test Random Copy Operator", "[core]") {
```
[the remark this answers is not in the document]
GumdropHorse: Thank you, resolved!

tests/tools/GenerativeWorld/Random.cpp
GumdropHorse: Thank you, resolved!

tests/tools/GenerativeWorld/Random.cpp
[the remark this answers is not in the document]
GumdropHorse: Thank you, resolved!

PR #32, opened 2026-02-16
Group 2

Cobathon:
Milestone Tests and Functional C++ classes

Included

- Tests for all 5 classes
- Further development within classes
- Proper File organization

### DIFF
--- a/source/tools/AgentAbility.cpp
+++ b/source/tools/AgentAbility.cpp
@@ -1,6 +1,0 @@
-/**
- * @file AgentAbility.cpp
- * @author lrima
- */
-
-#include "AgentAbility.hpp"


tests/tools/TestActionMap.cpp
AneeshJ10: Feedback

Testing and Error Checking:
You did a good job of testing the functions and edge cases of your code. You used std::optional well for error handling and assertions. 

Suggestions:
You tested most public member functions, but I didnt see any for SetDescription or clear. So maybe include those functions in your test cases.

source/tools/BehaviorTree/BehaviorTree.cpp
aryaplayzz: Hi!

So I read through your behavior tree implementation, and the algorithm looks great.

I had trouble running your test file. I tried running `make test` in the repo terminal, but I ran into compilation problems. I think this could be because our machines are different, the Makefile is set up in a way that does not work on my machine, or honestly, it could just be me doing something wrong on my end.

One thing I want to suggest is adding documentation within the code, especially since these are considered tools.

For example, having a quick description of what the file is about, what it does, and how to use it (since other people are going to be using these tools) would be really helpful for developers.

You could even have AI help write some of the documentation stuff (that is what I did, and I think it is better to have than not have).

I personally like a lot of comments in my code. I would rather read comments on what the code does than try to read the code and figure it out myself (that is more of a personal preference).

Anyways, back to the test files. I think right now, because everyone set up their test cases differently, it is hard to run them on other people’s end. Once we use a standard compilation setup like CMake (or something the whole company will use), running the tests will be much easier. I also think it will be easier after we have pushed all our code to the main branch and then go from there.

But regardless, for right now, could you let me know how you ran your tests? A simple “I did this and this” would be enough for me to figure out the rest. I am not asking for a whole tutorial or anything.

source/tools/BehaviorTree/BehaviorTree.hpp
aryaplayzz: See BehaviorTree.cpp for comments!

source/tools/BehaviorTree/BehaviorTree.cpp
```diff
@@ -0,0 +1,232 @@
+#include "BehaviorTree.hpp"
+#include <algorithm>
+#include <utility>
+
+namespace Pathfinding {
+namespace BehaviorTrees {
+
+
+bool Blackboard::Has(const std::string& key) const {
+    return data.find(key) != data.end();
```
aryaplayzz: Use data.contains(key)

simple code but urs still works

tests/tools/TestActionMap.cpp
AB123456kkkkk: For testing and error checking, all of the public member functions including constructors have unit tests. Used std::optional, and had errors.

source/tools/MemoFunction.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * @author Henry Finfrock
+ *
+ * This class acts as a function wrapper, caching outputs to improve performance
+ * Initial implementation, subject to change
+ */
+
+#pragma once
+
+#include <unordered_map>
+#include <map>
```
murra381: Very very small nitpick but I think in class Dr. Ofria mentioned commenting what you used each include for. I'm gonna add it into my code as well.

source/tools/MemoFunction.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace cse498 {
+
+ // Template for return and input types
+ template <typename R, typename... I>
+
+ class MemoFunction {
+ private:
+  // Key type so multiple inputs can be used as a cache key
+  using Key = std::tuple<I...>;
```
murra381: I really like that you are using I... throughout your code. It is nice to see a class concept in action. Really good job here.

source/tools/MemoFunction.hpp
```diff
@@ -0,0 +1,78 @@
...
+  std::map<Key, R> mCache;
+
+  // Queue that records order of caching, so when the cache reaches its
+  // limit it will remove the oldest
+  std::deque<Key> mQueue;
+
+  // The function being wrapped
+  std::function<R(I...)> mFunction;
+
+  // Size limit for cache (0 and less being no limit)
```
murra381: std::optional<size_t> mLimit = std::nullopt;
Would be a great alternative to this and another use of class concepts that would fit really well in my opinion. You would also be able to create a removelimit function if needing to disable the limit. Would be less confusing.

source/tools/MemoFunction.hpp
```diff
@@ -0,0 +1,78 @@
...
+   if (mLimit > 0 && CacheSize() > mLimit) {
+    Key oldest = mQueue.front();
+    mQueue.pop_front();
+    mCache.erase(oldest);
+   }
+
+   return output;
+  }
+
+  [[nodiscard]] size_t CacheSize() const { return mCache.size(); }
```
murra381: Good use of [[nodiscard]]

source/tools/MemoFunction.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+  // The function being wrapped
+  std::function<R(I...)> mFunction;
+
+  // Size limit for cache (0 and less being no limit)
+  size_t mLimit = 0;
+
+ public:
+  // Constructor, instantiated as MemoFunction<R, I1, I2, ... , IN> name(function)
+  explicit MemoFunction(std::function<R(I...)> mFunction) : mFunction(mFunction) {}
```
murra381: I would recommend changing mFunction (parameter name) to something different just so you don't confuse the two.

tests/tools/TestMemoFunction.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+TEST_CASE("MemoFunction: No argument", "[Memofunction]")
+{
+    MemoFunction<vector<char>> mf5(hello);
+    auto v = mf5();
+    REQUIRE(string(v.begin(), v.end()) == "hello");
+    REQUIRE(mf5.CacheSize() == 1);
+}
+
+int exception() { throw std::logic_error("test"); return 1; }
```
murra381: Good Error handling here!

tests/tools/TestMemoFunction.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    auto time1 = duration_cast<microseconds>(end1 - start1).count();
+    auto time2 = duration_cast<microseconds>(end2 - start2).count();
+    REQUIRE(time1 > time2 * 100);
+
+    //std::cout << "Time 1:" << time1 << std::endl;
+    //std::cout << "Time 2:" << time2 << std::endl;
+}
+
+TEST_CASE("MemoFunction: Cache limit", "[MemoFunction]")
```
murra381: I would like to see a few more examples of testing your setlimit here. What happens if I did SetLimit(-1)?

tests/tools/TestMemoFunction.cpp
```diff
@@ -0,0 +1,147 @@
+/**
+ * @file TestMemoFunction.cpp
+ * @author Henry Finfrock
+ *
+ * Catch2 unit tests for MemoFunction.
+ */
+
```
murra381: Unit Testing: All major functions tested (and constructor).

Edge Cases: Would like to see some more tests for SetLimit that directly cover some of the other edge cases.

Assertions: No assert statements in test file

Error Handling: The exception test was good. When I tested SetLimit myself and added a SetLimit(-1) it still passed all the test cases. I recommend adding some functionality to handle a case like this. I gave an example in the hpp file.

murra381 (commented):
Here is your review for MemoFunction. Overall, I liked it. Excited to see the full implementation of it.

source/tools/WorldPath.cpp
```diff
@@ -0,0 +1,223 @@
+/**
```
JordanSebagh: Although this problem is not directly in this class, it will affect your code. The  WorldPosition code appears to have missing structure and incomplete functions (ex: Set doesn’t set y and the braces look mismatched). There is not even a defined class in that file.

source/tools/WorldPath.hpp
```diff
@@ -0,0 +1,71 @@
...
+        double Length() const;
+
+        bool SelfIntersects() const;
+
+        // Returns indices of the two furthest points in the path.
+        std::optional<std::pair<std::size_t, std::size_t>> FurthestPointPair() const;
+        /**
+         * Reverses the world path in place.
+         */
+        WorldPath& reverse()
```
JordanSebagh: Most names are clear but there are a few inconsistencies with your naming conventions. reverse() and extend() are lowercase while most other methods are PascalCase. Try to pick one and be consistent.

JordanSebagh: reverse() should be noexcept if it truly can’t throw

source/tools/PathGenerator.hpp
gudimaya: Firstly, I apologize for the late review, I was very busy this weekend and was away from my laptop. 

For the ANode struct, I believe there should be a constructor for the ANode struct, mainly because I think the A Star search you have will be invalid because you have calls to ANode.

gudimaya: Im curious if constexpr is valid here since constexpr is compile time run, whereas the std::sqrt is not really compile time? However, I may be wrong.

gudimaya: I think these includes need to be added. 

#include <cmath>
#include <algorithm>
#include <limits>
#include <cassert>
#include <array>

source/tools/PathGenerator.cpp
gudimaya: Might need to include this library

#include <numbers>

source/tools/MemoFunction.hpp
LimEddie19: Great class with consistent coding style that looks like clang format. The variables. function names and parameters have great naming choices that make it easy to understand the inputs, outputs, and functionality of each function. Just like the previous commenter said I liked the use of the parameter pack. You could also use 'auto' instead of 'template' with the parameter pack and it would still be usable (this is just up to you whether you prefer template or auto). I agree with all the small changes the previous reviewer mentioned so I will not comment on those changes to avoid overlap. There will be other suggestions I will comment in between other lines of code.

Also, in group1 there is a class called RobinHoodMap that is a high-performance hash map using open addressing with robin hood hashing. It makes lookups and insertions significantly faster compared to <unordered_maps>. Although we have some time before we work out the integration between groups and their classes. This is something I would suggest improving the performance of your code. Great job!

tests/tools/TestMemoFunction.cpp
LimEddie19: Unit Testing: All basic tests created passed including the constructor.

Edge Cases: Would suggest incorporating cases mentioned by the previous Reviewer(murra381) and others like: SetLimit(0); or SetLimit(MAX_INT); I will also provide another suggestion in the testing below.

Assertions: You can test invalid inputs of SetLimit() and use assertions. For int you can test char or string inputs.

Error Handling: Great exception throw.

source/tools/PathGenerator.hpp
ZodiacalBee: I'm not sure what is happening to these comments?

source/tools/ActionMap/ActionMap.hpp
```diff
@@ -0,0 +1,329 @@
...
+    /**
+     * Helper to define what counts as a valid name.
+     * For now, it is only required to be non-empty.
+     *
+     * @param name attempting to be entered
+     * @return whether name is valid
+     */
+    static bool IsValidName(std::string_view name)
+    {
+        return !name.empty();
```
AneeshJ10: In the future, you could update this to include more strict name criteria based on the company standards

source/tools/WorldPath.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <cstddef>
+#include <iosfwd>
+#include <algorithm>
+#include <optional>
+#include <span>
+#include <utility>
+#include <vector>
+#include "../core/WorldPosition.hpp"
+
```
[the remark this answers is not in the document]
Luke-Antone: agree, I'll add it

source/tools/WorldPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include <limits>
+#include <ostream>
+
+namespace cse498
+{
+
+namespace
+{
+// Epsilon used for floating-point comparisons in geometry utilities.
+constexpr double kEps = 1e-9;
```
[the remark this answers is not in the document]
Luke-Antone: I'll stadardize it

source/tools/BehaviorTree/BehaviorTree.cpp
nahumj: Why didn't you follow the advice?

tests/tools/TestBehaviorTree.cpp
```diff
@@ -0,0 +1,785 @@
...
+ */
+
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../../source/tools/BehaviorTree/BehaviorTree.hpp"
+
+#include <vector>
+#include <string>
+
+using namespace Pathfinding::BehaviorTrees;
```
nahumj: Again, wrong namespace.

tests/tools/TestBehaviorTree.cpp
nahumj: Other than the namespace issue, the test cases look good.

tests/tools/TestWorldPath.cpp
nahumj: You don't test the "backwards-compatible" functions. They should either be removed entirely, or also part of the test suite.
I don't see any tests of fractional (non-integer) World positions.

source/tools/PathGenerator.cpp
nahumj: There are a lot of very complicated private functions in this file that should certainly be exposed and tested.

source/tools/ActionMap/ActionMap.hpp
```diff
@@ -0,0 +1,332 @@
+/**
+ * @file ActionMap.hpp
+ * @author Ty Maksimowski
+ *
+ * This class serves as a command registry that maps string action names
+ * to callable functions.
+ *
+ * AI Assistance:
+ * Portions of this implementation were developed with assistance from AI tools
```
nahumj: Any use of AI should be cited with a link to the conversation.

PR #36, opened 2026-02-17
Group16

MouseCR:
Group 16s base classes added into source/tools, test files in source/tests/tools. Classes added are ActionLog, DataLog, OutputManager, ReplayDriver, and Timer.


source/tools/ActionLog.hpp
GumdropHorse: Hi! Here is my peer review for the ActionLog class! Overall it is a well written class, though I do have some suggested changes and improvements listed below.


**Class Design**
Informative identifiers.
Identifiers are clear and easy to understand.

Avoids “magic numbers”.
There are no magic numbers in the code.

Intuitive and functional interface.
The class exposes only the necessary functionality. It is easy to use correctly and hard to misuse. Member variables are private. There are no default arguments, however they are classes that it doesn’t really make sense for them to have default arguments. Functions are not misleading.

Informative comments.
ActionLog.hpp has a good amount of comments that describe what things to without being overbearing. However, I would recommend moving the comment for LogAction to be above the class description. I almost didn’t see it because the line was so long.
ActionLog.cpp does not have any comments. I would recommend adding a few to help explain what exactly is happening, particularly in GetActionRange() and GetAgentActions().

Consistent style. 
Style choices are consistent. Classes use PascalCase and variables use snake_case. However, I would recommend updating the member variables to have the prefix m_, to make it more clear in the code when something is a member variable as opposed to a regular variable.

Visual appearance. 
Overall, the code is pretty easy to scan through and is visually well structured. I would note however that the definition for LogAction() line is a bit long.


**Implementation**
No warnings during compilation.
There were not any warnings or errors during compilation.

Proper uses of namespaces.
None of the code is within the namespace cse498{}. I would recommend changing this. However, the class does not pollute the global namespace at all.

Adheres to the C++ standard. Does the code avoid undefined behavior? If possible, make sure the code compiles and runs properly on different compilers and different platforms.
.The code properly avoids undefined behavior.

Appropriate Modularity.
Unnecessary code duplication is avoided.

Uses modern C++ constructs.
One recommendation I have, is to update GetActions(), GetActionRange(), GetAgentActions() and GetActionCount() to all use std::expected, as this would help with error handling that I do not see the functions addressing.

Uses standard library algorithms.
Not applicable, the code has no use for standard library algorithms.

Pointer handling.
Not applicable, pointers are not used and not needed.

const correctness.
Const correctness is maintained throughout the code.


**Testing and Error Checking**
Unit testing.
The unit testing does not make use of the Catch2 testing. I would highly recommend updating the tests to use the Catch2 framework demonstrated in class.  Other than that though, the tests are clearly labeled. All public functions are tested, and they trigger all private portions of the class’ code.
I would also like to note that two of the tests in TimerTest did not pass, however they are not for this class so I will not focus on it for this review unless you have more questions about the exact failure messages I got.

Edge cases.
Overall, the test cases do cover most edge cases. However, I would recommend also testing for a large amount of actions stored in the action log.

Assertions.
Asserts exist in the test file, but I would also recommend adding some to ActionLog.cpp.

Error handling.
There are no error conditions that are checked for. I would recommend adding checks to GetActionRange() to make sure that the input is valid, and I would recommend the same with GetAgentActions().

source/tools/DataLog.cpp
```diff
@@ -0,0 +1,150 @@
...
+ **/
+
+#include "DataLog.hpp"
+#include <vector>
+#include <cstddef>
+#include <optional>
+#include <string>
+#include <utility>
+#include <algorithm>
+#include <chrono>
```
AbbyFranz: Appropriate Modularity- The includes in the cpp file are code duplication because they are already included in the hpp file, and you only need them in one spot.

tests/tools/DataLogTest.cpp
AbbyFranz: Unit testing - Just for clarity I'd mention that your positive test shows that you test finding median for even indexing, and that your negative test shows that you test finding median for odd indexing. This will just help explicitly show that you tested them since they dont have their own named tests.

source/tools/DataLog.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * @author Aneesh Joshi
+ * @note Status: PROPOSAL
+ * 
+ * The goal of this class is to provide a time based sequence of numeric values and provides statistics on them.
+ * Samples are stored in the format of (Value, timestamp/seconds since start)
+ * Caller will add a numeric value and the class will associate a timestamp from when the instance was constructed.
+ **/
+
+#include "DataLog.hpp"
+#include <vector>
```
Lak453: No need to include it again because it has already been declared in the header file.

source/tools/OutputManager.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * Verbose: More detailed than Info, but less than Debug.
+ * Info: Confirmation that things are working as expected.
+ * Warn: An indication that something unexpected happened, or indicative of some problem in the near future.
+ * Error: Due to a more serious problem, the software has not been able to perform some function.
+ * Silent: No logging output.
+ */
+
+enum class LogLevel : uint8_t 
+{
+    DEBUG = 0,
```
ZhixiangMiao: As described above, you should name "DEBUG" as "Debug" to have a consistent style.

tests/tools/OutputManagerTest.cpp
ZhixiangMiao: As I mentioned, you should also add your GetMinLogLevel() to test if it works, and your LevelName function to test if it returns the correct level name.

source/tools/OutputManager.cpp
```diff
@@ -0,0 +1,207 @@
...
+    std::scoped_lock lock(m_mutex); //Lock mutex
+    m_csvEnabled = enabled;
+}
+
+/*@brief Sets the CSV output file path
+ * @param path File path to write CSV rows to
+ * @param append True to append, false to overwrite
+*/
+
+void cse498::OutputManager::SetCsvPath(const std::string& path, bool append)
```
ZhixiangMiao: You should check if the file and path are valid. If it's not, you should throw an error for error handling.

source/tools/OutputManager.hpp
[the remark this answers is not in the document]
murra381: Thank you so much for the feedback. This was very informative. I will get back to you with relevant changes such as getting rid of the duplicate include, testing GetMinLogLevel, and making my style more consistent. I appreciate it.

source/tools/DataLog.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    const auto size = stored_data.size();
+    const auto midpoint = size/2;
+
+    //check if the size of the data log is odd or even to calculate the median
+    if(size % 2 == 1){
+        return stored_data[midpoint];
+    }
+    
+    double median = (stored_data[midpoint] + stored_data[midpoint - 1]) / 2.0;
```
Lak453: Consider adding a short comment clarifying that midpoint is size/2 and you average the two middle values for even-sized datasets.

source/tools/Timer.hpp
```diff
@@ -0,0 +1,45 @@
...
+        ~Timer() = default;
+        void start(std::string timerName);
+        void stop(std::string timerName);
+        void restart(std::string timerName);
+        double getTime(std::string timerName);
+
+        /**
+         * Returns whether the timer with the provided name is currently running. If the timer doesn't exist, returns false.
+         */
+        bool isRunning(std::string timerName) { return mTimers.find(timerName) != mTimers.end() ? mTimers.at(timerName).isRunning : false; }
```
yhenry6: You can mark this function const and take const std::string& (or std::string_view) so it can be called on const objects and avoids copying.

source/tools/Timer.hpp
```diff
@@ -0,0 +1,45 @@
+/**
+ * @file Timer.hpp
+ * @author Nathan Charters
+ *
+ * A Timer class that can be used to measure precise time intervals.
+ */
+
+#ifndef TIMER_H
```
paulbui081: Proper uses of namespaces: all code is within namespace cse498 {...} and avoids 'using namespace std'

source/tools/Timer.hpp
```diff
@@ -0,0 +1,45 @@
...
+    public:
+        Timer();
+        ~Timer() = default;
+        void start(std::string timerName);
+        void stop(std::string timerName);
+        void restart(std::string timerName);
+        double getTime(std::string timerName);
+
+        /**
+         * Returns whether the timer with the provided name is currently running. If the timer doesn't exist, returns false.
```
paulbui081: I would recommend moving the implementation of isRunning(...) into Timer.cpp for better structural clarity. Having the comments and implementation split across the header and source files could be confusing for users and make the interface harder to follow.

tests/tools/TimerTest.cpp
```diff
@@ -0,0 +1,88 @@
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+#include <thread>
+
+#include "../../source/tools/Timer.hpp"
+
+TEST_CASE("Test Timer Constructor and Basic Functionality", "[tools]")
+{
```
paulbui081: Test cases pass and are overall very well-documented. Each test case explains the intended results, being brief and easy to understand. Unit tests are present for all public member functions and are well organized, good job!

source/tools/ReplayDriver.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef REPLAYDRIVER_H
+#define REPLAYDRIVER_H
+
+#include "ActionLog.hpp"
+
+#include <string>
+#include <vector>  
+#include <fstream>
+
+
```
joeyhyun1: Not using namespace cse498

source/tools/ReplayDriver.cpp
joeyhyun1: Overall the code is good and compiled on my windows machine. There are just a few things in terms of changes that I proposed and those can be seen more specifically in the actual code. First there are a few inconsistencies with the naming in the code. For example using PascalCase vs camelCase for variable names and Function names. There is also some inconsistencies with the spacing in front of if statements. As for the test file all the tests pass and it seems to be extensive enough.

source/tools/ReplayDriver.cpp
```diff
@@ -0,0 +1,103 @@
...
+// Replays actions at intervals of 'step' (e.g., every 10th action)
+void ReplayDriver::ReplayByStep(int step) { 
+    if(IsActionLogSet() == true && step > 0) {
+        lastReplayedActions.clear(); // Clear previous replayed actions
+        for(int i = 0; i < m_actionLog->GetActionCount(); i += step) {
+            Action action = m_actionLog->GetActions()[i];
+            verifyAction(action);
+        }
+    }
+}
```
lunncurt: if step <= 0, function does nothing. Possible return a bool to verify result

lunncurt (commented):
ReplayDriver has no glaring issues and is well made. I left a few comments (see above) about pointer type, possible stale state may benefit from different return type, unnecessary copies, and also noticed a lack of cse498 namespace. Overall well done! Comments pointed out small things, not breaking issues.

CShon (commented):
I forgot to actually submit the review, so you may not have been able to see it until now. Really sorry about that, hope it didn't cause any issues!

source/tools/Timer.cpp
```diff
@@ -0,0 +1,100 @@
+/**
+ * @file Timer.cpp
+ * @author Nathan Charters
+ */
+
+#include "Timer.hpp"
+
+double NANOSECONDS_IN_SECOND = 1000000000.0;
```
nahumj: This should be constexpr (but this mistake was made after peer review).

tests/tools/DataLogTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+/*
+Test checks functions of the datalog with no samples
+*/
+TEST_CASE("Empty Log Test", "[tools]"){
+
+    cse498::DataLog log;
+    CHECK(log.Count() == 0);
+
+    CHECK(log.DataSamples().empty());
+    CHECK(log.DataSamples().size() == log.Count());
```
nahumj: What is the point of this check?

tests/tools/DataLogTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    CHECK(log.Count() == 1);
+    CHECK(log.DataSamples().size() == log.Count());
+    CHECK(log.DataSamples()[0].first == Approx(10.1));
+
+    CHECK(log.Median().value() == Approx(10.1));
+    CHECK(log.Mean().value() == Approx(10.1));
+    CHECK(log.Min().value() == Approx(10.1));
+    CHECK(log.Max().value() == Approx(10.1));
+
+    log.Clear();
```
nahumj: Again, the checks on clear shouldn't be here.

source/tools/OutputManager.cpp
```diff
@@ -0,0 +1,272 @@
...
+ * @param message The message to log
+*/
+void cse498::OutputManager::Log(LogLevel level, LogCategory category, std::string_view message) 
+{   
+    bool time_stamp_enabled = true;
+    std::chrono::steady_clock::time_point start_copy;
+
+    {
+        std::scoped_lock lock(m_mutex);
+        if (m_min == LogLevel::Silent || static_cast<uint8_t>(level) < static_cast<uint8_t>(m_min))
```
nahumj: My have a Silent level that can never be used?

source/tools/ReplayDriver.hpp
nahumj: This doesn't match the proposed spec; it doesn't take control of agents and make them perform actions in any way.

source/tools/Timer.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (mTimers.find(timerName) != mTimers.end())
+    {
+        if (mTimers.at(timerName).isRunning)
+        {
+            // If the timer exists, return the accumulated duration plus
+            // the current running time if it's still running.
+            auto startTime = mTimers.at(timerName).startTime;
+            auto currentTime = std::chrono::steady_clock::now();
+            auto elapsed = mTimers.at(timerName).accumulatedDuration + (currentTime - startTime);
+            return elapsed.count() / 1000000000.0; // Convert from nanoseconds to seconds.
```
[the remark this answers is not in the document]
ncharters: Yeah, I completely missed that, I've added a global variable at the top of the file that holds the conversion rate from nanoseconds to seconds.

source/tools/Timer.hpp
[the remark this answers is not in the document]
ncharters: Yeah, originally my intention for the class was to have it be just a single timer, but then I pivoted to creating something more akin to a container for timers in the hopes that maybe I could create functions for comparisons between timing instances in the future, and I didn't quite think of renaming the file to reflect that. I really like your idea of creating two classes, one that is the Timer object with a start functionality, stop functionality, etc. and another that is the container for those objects where I can implement aggregate functions or comparative. I think that makes a lot more sense intuitively and will relieve a bit of the confusion surrounding the functionality. If I go through with that change, I will likely not use the TimerInfo struct at all, and that information will just be member variables within the Timer object class.

tests/tools/TimerTest.cpp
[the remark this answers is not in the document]
ncharters: Ok, I talked to you in class regarding this particular issue and I'll get on fixing that so that my test cases can be run with the rest of the testing suite.

source/tools/OutputManager.hpp
```diff
@@ -0,0 +1,179 @@
...
+        * @param enabled True to enable timestamps, false to disable
+        */
+        void EnableTimestamps(bool enabled);
+
+        /*@brief Logs a message at a given level and category
+        * @param level The log level of the message
+        * @param category The log category of the message
+        * @param message The message to log
+        */
+        void Log(LogLevel level, LogCategory category, std::string_view message);
```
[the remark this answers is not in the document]
murra381: I have now added defaults, thanks for catching that. It now defaults to Info and System. There is also a Log("message") convenience overload.

source/tools/OutputManager.hpp
```diff
@@ -0,0 +1,179 @@
...
+        * @param append True to append, false to overwrite
+        */
+        void SetCsvPath(const std::string& path, bool append = true);
+
+    private:
+        mutable std::mutex m_mutex; /// Mutex for thread operations
+        LogLevel m_min{LogLevel::Info}; /// Minimum log level
+
+        bool m_timestamps{false}; /// Flag for timestamps
+        std::chrono::steady_clock::time_point m_startTime; /// Start time
```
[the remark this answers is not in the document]
murra381: I updated my classes to use the Timer class and then switched it to the newly updated Stopwatch class. It uses the internal stopwatch to generate timestamps now as opposed to the old, local one in my class.

PR #37, opened 2026-02-17
Group 18's fully functional C++ classes + Video Overview

Mehver:


source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,376 @@
...
+  height_ = height_px;
+  if (!element_.isNull()) {
+    if (width_px > 0) {
+      element_["style"].set("width", ToPx(width_px));
+    }
+    if (height_px > 0) {
+      element_["style"].set("height", ToPx(height_px));
+    }
+    // Stretch to exact dimensions (no aspect ratio preservation)
+    element_["style"].set("objectFit", std::string("fill"));
```
ZodiacalBee: This function SetSize and Resize seem very redundant and I don't see why. I would add an extra argument as an enum class with default argument of "FILL" and otherwise the otther options: Contain, Fill, cover,...

source/Interfaces/WebUI/WebImage/WebImage.hpp
```diff
@@ -0,0 +1,150 @@
...
+  /// scaled to fit within the given bounding box while preserving its
+  /// original aspect ratio (CSS object-fit: contain). Otherwise it
+  /// stretches to fill the exact dimensions (CSS object-fit: fill).
+  /// @param width_px              Target width in pixels (must be > 0)
+  /// @param height_px             Target height in pixels (must be > 0)
+  /// @param maintain_aspect_ratio If true, scale to fit; if false, stretch.
+  void Resize(int width_px, int height_px, bool maintain_aspect_ratio = false);
+
+  /// Get the display width.
+  int GetWidth() const;
```
ZodiacalBee: add the [[nodiscard]] attributes to these functions. It is a good habit I think

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,376 @@
...
+  element_.set("id", id_);
+  element_.set("src", src_);
+  element_.set("alt", alt_text_);
+  doc["body"].call<void>("appendChild", element_);
+
+  AttachListeners();
+}
+
+WebImage::~WebImage() {
+  if (!element_.isNull()) {
```
ZodiacalBee: Could move this into a separate function "CleanupElement" or something and so you can use it in operator=(WebImage&& other) where it can also be substituted. Your choice though because this is small-ish

tests/Interfaces/WebUI/WebImageTest.cpp
ZodiacalBee: Seems like thorough testing. I hope most of this is generated by AI because it seems like a pain otherwise. I'm not sure why it is functions then calling those functions with the test cases.. it seems redundant. 

You tested most things that I could imagine.

source/Interfaces/WebUI/WebCanvas/WebCanvas.hpp
```diff
@@ -0,0 +1,76 @@
...
+    void renderFrame();
+
+    // ---- Immediate-mode primitive drawing (whiteboard-level) ----
+    // Notes:
+    // - In non-Emscripten builds these are safe no-ops (useful for unit tests).
+    // - In Emscripten builds these issue Canvas2D calls on the <canvas id="Id()">.
+    void Clear(const std::string& cssColor = "#000000");
+
+    void DrawLine(float x1, float y1, float x2, float y2,
+                  float lineWidth = 1.0f,
```
sitarakb: Avoids "magic numbers": I noticed in DrawLine, DrawCircle, & DrawPolygon ,lineWidth = 1.0f is used, and in DrawPoint, radius = 2.0f is used. In order to use constant variables instead of arbitrary numbers throughout your code, I would suggest setting constants for these numbers. Other than that, looks good!

source/Interfaces/WebUI/WebCanvas/WebCanvas.hpp
```diff
@@ -0,0 +1,76 @@
...
+                   const std::string& fillColor = "#ffffff");
+
+    // If fillColor is empty => no fill.
+    void DrawPolygon(const std::vector<Vec2>& points,
+                     const std::string& strokeColor = "#ffffff",
+                     float lineWidth = 1.0f,
+                     const std::string& fillColor = "");
+
+private:
+    std::vector<std::unique_ptr<ICanvasElement>> m_elements;
```
sitarakb: Intuitive and functional interface: all member variables are private, there is a clear separation of functions based off functionality & it seems hard to misuse due to no misleading functions!

source/Interfaces/WebUI/WebCanvas/WebCanvas.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "../internal/IDomElement.hpp"
+#include "../internal/ICanvasElement.hpp"
+
+/**
+ * WebCanvas
```
sitarakb: Informative comments & visual appearance: great jobs on your comments! the class documentation does a good job at explaining its duality, and the separation of functions is clear. This made your code easy to scan with its structure.

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,298 @@
...
+// Notes:
+// - Do NOT define CATCH_CONFIG_MAIN here if another test file already defines it.
+
+#include <type_traits>
+#include <memory>
+#include <vector>
+#include <string>
+#include <utility>
+
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
```
sitarakb: No warnings during compilation: unfortunately, I was unable to run your group's tests due to this error: fatal error: 'emscripten/val.h' file not found
    6 | #include <emscripten/val.h>
I believe this is because emscripten needs to be installed on all devices running your code. For reference, my team created a install target that installs our GUI library on any device running the command.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+    // For free layout, apply positioning relative to root_ if set
+    if (type_ == LayoutType::Free) {
+      int top = elem->FreeTop();
+      int left = elem->FreeLeft();
+      val est = el["style"];
+      if (top >= 0 || left >= 0) {
+        est.set("position", std::string("relative"));
+        if (top >= 0) {
+          est.set("position", std::string("relative"));
```
Cobathon: Duplicate Position that I believe can be removed

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -0,0 +1,392 @@
...
+/// Sets the gap/spacing between child elements.
+void WebLayout::SetSpacing(int pxv) noexcept {
+  assert(pxv >= 0);
+  spacing_ = pxv;
+}
+
+/// Adds a child element to this layout.
+/// The element is appended to the root's DOM node and tracked for later Apply()
+/// calls.
+bool WebLayout::AddElement(IDomElement* elem, Alignment align) noexcept {
```
Cobathon: I don't see any code that disallows duplicate, I'm not sure if that's intended or not.

Cobathon (commented):
Overall, this is well-structured and well-written code, It's clean and readable with only minor cleanup and adjustments needed.

tests/Interfaces/WebUI/WebCanvasTest.cpp
Mehver: I recall that [the Week 3 course materials](https://emscripten.org/docs/getting_started/index.html) provided in [the Schedule page](https://cse498.github.io/schedule.html) include instructions on setting up the Emscripten environment. Of course, if necessary, the company can address this more specifically when configuring the build toolchain later.

source/Interfaces/WebUI/WebCanvas/WebCanvas.hpp
Mehver: Fixed in https://github.com/CSE498/Spring2026-CompanyA/commit/115d7f497f99397394564938ba84599c3dacc32a

source/Interfaces/WebUI/WebButton/main.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <iostream>
+#include <emscripten.h>
+
+WebLayout* layout = nullptr;
+WebButton* btn1 = nullptr;
+WebButton* btn2 = nullptr;
+WebButton* btn3 = nullptr;
+
+int main() {
+  layout = new WebLayout("button-demo");
```
Luke-Antone: posible memory leak, could use RAII to prevent this

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -0,0 +1,706 @@
...
+ }
+
+    virtual ~MockDomElement() = default;
+
+    void mountToLayout(WebLayout& /*parent*/,
+                       Alignment /*align*/) noexcept override {
+        mountCount++;
+    }
+
+    void unmount() noexcept override {
```
Cobathon: mount() and unmount() do not throw, but it does not validate lifecycle effects (e.g., DOM presence or mount counters in MockDomElement). Adding assertions on side effects would strengthen this test.

source/Interfaces/WebUI/WebButton/WebButton.hpp
```diff
@@ -0,0 +1,66 @@
...
+  std::string label_;
+  std::function<void()> callback_;
+  bool is_enabled_ = true;
+  bool is_visible_ = true;
+  int width_ = 0;
+  int height_ = 0;
+  std::string bg_color_;
+  std::string text_color_;
+  emscripten::val element_;
+  std::string id_;
```
marksandychez: These member variables could be explicitly declared to their default initial values here

marksandychez: To denote these as member variables, consider switching over to the "m" prefix and removing the trailing underscore. Votes in the discord seem to lean that this will be the new convention for the company as of today.

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,240 @@
...
+  width_ = width;
+  height_ = height;
+  if (!element_.isNull()) {
+    if (width > 0) {
+      element_["style"].set("width", ToPx(width));
+    }
+    if (height > 0) {
+      element_["style"].set("height", ToPx(height));
+    }
+  }
```
marksandychez: Here, if height or width == 0, it will pass the asserts, set the member variable(s) to 0, but not actually change the element height or width in the if blocks below. This might cause bugs where member variables do not match the visual output on the screen if I am interpreting this correctly.

source/Interfaces/WebUI/WebButton/main.cpp
marksandychez: This appears to be a demo file, consider storing this locally instead of on the repository to avoid confusion with what is critical to the WebButton class

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -0,0 +1,392 @@
...
+  root_.set("id", id_);
+  val style = root_["style"];
+  style.set("boxSizing", std::string("border-box"));
+  document_["body"].call<void>("appendChild", root_);
+  assert(!root_.isNull() && !root_.isUndefined());
+  assert(root_["id"].as<std::string>() == id_);
+}
+
+/// Destructor removes the root element from the DOM and cleans up references.
+WebLayout::~WebLayout() noexcept {
```
Deni-Tepic: Not sure about this but ...
 
If a WebLayout is mounted inside another layout, the destructor removes its DOM node but doesn’t notify the parent layout. Could that leave a stale pointer in the parent’s children_ vector?

source/Interfaces/WebUI/WebTextbox/WebTextbox.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+  // RectPx is just a lightweight struct to hold the pixel-based bounding rectangle of a DOM element.
+  struct RectPx {
+    double x = 0.0;
+    double y = 0.0;
+    double w = 0.0;
+    double h = 0.0;
+  };
+
+  /* ----- Constuctor / Destructor / Move semantics ----- */
```
tmaks2: Small typo with "Constuctor" -> Constructor

source/Interfaces/WebUI/WebTextbox/WebTextbox.hpp
```diff
@@ -0,0 +1,103 @@
...
+  std::string text_;
+  bool is_visible_ = true;
+
+  // font/fallback behavior
+  std::string requested_font_family_;
+  std::string fallback_font_family_ = "sans-serif";
+
+  // basic style state
+  float font_size_px_ = 16.0f;
+  float line_height_px_ = 0.0f; // 0 let browser choose
```
tmaks2: You say this lets browser choose when set to 0, but SetLineHeight() asserts >0. I think you should either allow 0 as a way to reset to normal or add something like a ClearLineHeight() helper

source/Interfaces/WebUI/WebTextbox/WebTextbox.cpp
```diff
@@ -0,0 +1,412 @@
...
+  ApplyStyles();
+}
+
+void WebTextbox::SetItalic(bool enabled)
+{
+  italic_ = enabled;
+  ApplyStyles();
+}
+
+void WebTextbox::SetAlignment(const std::string& alignment)
```
tmaks2: Using std::string is easy, but can also be easy to misuse cause of typos. Maybe consider using an enum TextAlign { Left, Center, Right }

tests/Interfaces/WebUI/WebImageTest.cpp
```diff
@@ -0,0 +1,629 @@
+#define CATCH_CONFIG_MAIN
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../../../source/Interfaces/WebUI/WebImage/WebImage.hpp"
+#include <string>
+
```
anaghaj9: You have to use using namepsace cse498 here I believe

tests/Interfaces/WebUI/WebImageTest.cpp
```diff
@@ -0,0 +1,629 @@
...
+TEST_CASE("Constructor sets source and alt text", "[WebImage]") { test_constructor(); }
+TEST_CASE("Constructor default alt text is empty", "[WebImage]") { test_constructor_default_alt(); }
+TEST_CASE("SetSource and GetSource", "[WebImage]") { test_set_get_source(); }
+TEST_CASE("SetSource resets loaded and error state", "[WebImage]") { test_set_source_resets_state(); }
+TEST_CASE("SetAltText and GetAltText", "[WebImage]") { test_set_get_alt_text(); }
+TEST_CASE("SetSize sets width and height", "[WebImage]") { test_set_size(); }
+TEST_CASE("SetSize handles zero dimensions", "[WebImage]") { test_set_size_zero(); }
+TEST_CASE("Resize without aspect ratio", "[WebImage]") { test_resize_no_aspect_ratio(); }
+TEST_CASE("Resize with aspect ratio", "[WebImage]") { test_resize_maintain_aspect_ratio(); }
```
anaghaj9: I would leave inline comments explaining what these test cases are since there are so many of them and it's not easy to understand what they do as a first time code reader. I would also break up the test cases like you did in the earlier ones to show differences in use cases

tests/Interfaces/WebUI/WebImageTest.cpp
```diff
@@ -0,0 +1,629 @@
...
+  CHECK_MSG(img.GetSource() == "second.png", "source should be second.png");
+
+  img.SetSource("third.png");
+  CHECK_MSG(img.GetSource() == "third.png", "source should be third.png");
+
+  img.SetSource("");
+  CHECK_MSG(img.GetSource().empty(), "source should be empty");
+  PASS();
+}
+
```
anaghaj9: Tests for multiple edge cases, and considering multiple different use cases - very thorough and done well! Unit testing for all public member functions is also done.

source/Interfaces/WebUI/WebImage/WebImage.hpp
```diff
@@ -0,0 +1,150 @@
...
+  std::string src_;
+  std::string alt_text_;
+  int width_ = 0;
+  int height_ = 0;
+  double opacity_ = 1.0;
+  bool is_visible_ = true;
+  bool is_loaded_ = false;
+  bool has_error_ = false;
+  ImageErrorMode error_mode_ = ImageErrorMode::BlankRect;
+  std::string placeholder_color_ = "#CCCCCC";
```
anaghaj9: I think placeholder_color is considered to a magic number, so ideally it should be named a constant something static constexpr const char* defaultPlaceHolderColor  = "#CCCCC"

source/Interfaces/WebUI/WebImage/WebImage.hpp
```diff
@@ -0,0 +1,150 @@
...
+  void Show();
+  /// Hide the image.
+  void Hide();
+  /// Check if the image is visible.
+  bool IsVisible() const;
+
+  // ----- Loading State & Error Handling -----
+
+  /// Mark the image as loaded or not (for tracking async loading).
+  void MarkLoaded(bool loaded);
```
anaghaj9: I'm confused on why you made this public, what was your thought process? I feel like this would maybe be a private function assuming the user doesn't manually mark an image as loaded

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,376 @@
...
+    var ptrVal = $1;
+    el.addEventListener("load", function() {
+      Module._WebImage_handleLoad(ptrVal);
+    });
+    el.addEventListener("error", function() {
+      Module._WebImage_handleError(ptrVal);
+    });
+  }, element_.as_handle(), reinterpret_cast<intptr_t>(self));
+}
+
```
anaghaj9: I suggest adding one-line comments quickly describing the function, because there are so many, it's easy to get lost in the code.

source/Interfaces/WebUI/WebCanvas/WebCanvas.hpp
finfroc3: Very good class. The formatting makes the whole class clean and readable, all naming is very intuitive (I found nothing confusing or redundant), and comments are used to great effect in organizing everything, making it very clear what things fall under each intended use of the class. While the functions and variables are generally understandable by themselves, I would still suggest including a few more explanatory comments just to cover all of your bases (things like the drawing functions, Vec2, m_elements and its related functions), though this is more of a subjective idea. Now that the company has come to an agreement on naming conventions, you should also make sure everything is in line with that. Otherwise, though, it is very well written and functional, and I don't have any other real objective critiques to give you. Good job!

source/Interfaces/WebUI/WebTextbox/WebTextbox.hpp
```diff
@@ -0,0 +1,103 @@
+#ifndef WEBTEXTBOX_HPP_
+#define WEBTEXTBOX_HPP_
+
+#include <functional>
+#include <string>
+#include <emscripten/val.h>
+
+#include "IDomElement.hpp"   // same base used by WebImage
+#include "WebLayout.hpp"     // for mountToLayout signature
```
kianamay: Informative identifiers: good all names are clear, understandable and consistent 
Avoids "magic numbers": The class uses clear default values like font_size_px_ = 16.0f and fallback_font_family_ = "sans-serif" instead of magic numbers
Intuitive and functional interface :all member variables properly private and with  default parameters. Member functions have a clearly defined purpose. 
Informative comments: Overall comments are helpful in understanding the code but more detail could be added on emscripten usage for those not familiar. 
Consistent style: Good! Will need to convert to for future reference: Code Style: Clang Format Variable name: camelCase (without m-prefix for member variables) TBD Function name: PascalCase Global Variables: FULL_UPPERCASE Libraries: lowercase
Visual appearance: The code is well-formatted!

source/Interfaces/WebUI/WebButton/main.cpp
```diff
@@ -0,0 +1,39 @@
...
+  layout->SetLayoutType(LayoutType::Vertical);
+  layout->SetSpacing(10);
```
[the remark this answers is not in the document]
hoangtu0701: I'll pull this out into a named constant

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,240 @@
...
+int WebButton::next_id_counter_ = 1;
+
+static val GetDocument() {
+  return val::global("document");
+}
+
+static std::string ToPx(int value) {
+  return std::to_string(value) + "px";
+}
+
```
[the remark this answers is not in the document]
hoangtu0701: Sure :D I kept comments minimal on purpose but I can add brief doc comments on the public functions in the header to describe inputs/outputs.

source/Interfaces/WebUI/WebButton/WebButton.cpp
hoangtu0701: Oh right. My intent was that 0 means "no explicit size set" so the button uses its default/auto size. But you're right that it's not obvious. I'll add a comment clarifying that behavior and consider whether 0 should reset the style back to auto instead of just skipping.

source/Interfaces/WebUI/WebButton/WebButton.hpp
hoangtu0701: Some of them already have defaults (width_ = 0, height_ = 0, is_enabled_ = true, etc.) but you're right that bg_color_, text_color_, element_, and id_ are relying on implicit defaults. I'll make them all explicit for consistency.

hoangtu0701: Thanks for the heads up on the convention change. I'll switch to the m prefix

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
alexelkins23: fixed

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
alexelkins23: fixed

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -0,0 +1,706 @@
...
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "../../../source/Interfaces/WebUI/WebLayout/WebLayout.hpp"
+#include "../../../source/Interfaces/WebUI/internal/IDomElement.hpp"
+
+// ---------------------------
+// Test Helpers & Stubs
+// ---------------------------
+
+/// A minimal test double for IDomElement that tracks lifecycle and state
+class MockDomElement : public IDomElement {
```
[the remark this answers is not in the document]
alexelkins23: fixed

source/Interfaces/WebUI/WebTextbox/WebTextbox.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+/// Manages an HTML text element (<div>) from C++ via Emscripten.
+/// Rendered as a DOM element via WebLayout.
+class WebTextbox : public IDomElement {
```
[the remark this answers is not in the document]
motowsk2: Added cse498 namespace

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,232 @@
...
+  if (!mElement.isNull()) {
+    mElement.set("textContent", mLabel);
+  }
+}
+
+std::string WebButton::GetLabel() const {
+  return mLabel;
+}
+
+void WebButton::SetCallback(std::function<void()> callback) {
```
karensuzue: I think passing `callback` by value and then doing `std::move` is a reasonable approach here. If we want to be a bit more efficient, we could add overloads for lvalues + rvalues. That way, passing an lvalue does a single copy, while the original function does a copy into the parameter followed by a move into mCallback.

So something like this:
```

void SetCallback(const std::function<void()> & cb) { mCallback = cb; }
void SetCallback(std::function<void()> && cb) { mCallback = std::move(cb); }
```

karensuzue: -R?

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,232 @@
...
+  return mIsVisible;
+}
+
+void WebButton::MountToLayout(WebLayout& parent, Alignment align) {
+  parent.AddElement(this, align);
+}
+
+void WebButton::Unmount() {
+  if (mElement.isNull()) return;
+
```
karensuzue: I think it's worth adding a comment explaining what happens after an unmount. My understanding is that mElement would still exist and be reused, it just disappears from the DOM. But someone might mistake this for setting mElement to null or freeing the node

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,232 @@
...
+  val parent = mElement["parentNode"];
+  if (!parent.isNull() && !parent.isUndefined()) {
+    parent.call<void>("removeChild", mElement);
+  }
+}
+
+void WebButton::SyncFromModel() {
```
karensuzue: This function definitely needs some documentation.

karensuzue: -C

source/Interfaces/WebUI/WebButton/WebButton.cpp
```diff
@@ -0,0 +1,232 @@
...
+void WebButton::HandleClick() {
+  Click();
+}
+
+void WebButton::AttachClickListener() {
+  if (mElement.isNull()) return;
+
+  EM_ASM({
+    var el = Emval.toValue($0);
+    var ptrVal = $1;
```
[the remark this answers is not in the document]
karensuzue: What you could do instead is set up a mapping/registry of sorts between string IDs and WebButton objects. That way, you can pass in the button's ID, then look up the corresponding address when WebButton_handleClick is called.

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
```diff
@@ -0,0 +1,162 @@
...
+    // Safe no-op placeholder for now.
+}
+
+// ---- Canvas content management ----
+void WebCanvas::AddElement(std::unique_ptr<ICanvasElement> element)
+{
+    if (!element) {
+        return;
+    }
+    m_elements.emplace_back(std::move(element));
```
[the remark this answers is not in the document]
karensuzue: -C

tests/Interfaces/WebUI/WebButtonTest.cpp
```diff
@@ -0,0 +1,294 @@
...
+  bool AddElement(const std::string&, Alignment = Alignment::Start) { return true; }
+};
+
+#define CHECK_MSG(cond, msg) \
+  do { INFO(msg); CHECK(cond); } while(0)
+
+TEST_CASE("Constructor sets label", "[WebButton]") {
+  WebButton btn("Click Me");
+  CHECK_MSG(btn.GetLabel() == "Click Me", "label mismatch");
+  CHECK_MSG(btn.IsEnabled() == true, "default should be enabled");
```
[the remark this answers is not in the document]
karensuzue: Ah, but hold on! If the width or height is 0, it reverts to "auto", which means btn.GetWidth() and btn.GetHeight() won't return the correct values. I'd be careful with that...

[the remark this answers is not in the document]
karensuzue: -C -R

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+    void Draw(WebCanvas& /*canvas*/) override {
+        ++drawCount;
+        order_.push_back(mId);
+    }
+
+    int drawCount = 0;
+
+private:
+    int mId;
+    std::vector<int>& order_;
```
karensuzue: What exactly is order_ and why is it a std::vector?

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+};
+
+struct LifetimeElement : public ICanvasElement {
```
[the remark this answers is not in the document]
karensuzue: Ah I understand what these classes are for now. Still, I think this would benefit from a few comments

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    LifetimeElement() { ++alive; }
+    ~LifetimeElement() override { --alive; ++destroyed; }
+
+    void Draw(WebCanvas& /*canvas*/) override {}
+
+    static void reset() { alive = 0; destroyed = 0; }
+};
+
+struct OrderElement : public ICanvasElement {
```
karensuzue: Again, I think a few comments explaining the purpose of these classes would be nice to have

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+// A minimal WebLayout stub for unit tests.
+// This avoids pulling in the real WebLayout implementation (which may depend on emscripten).
+class WebLayout {};
+
+// ---------------------------
+// Test helpers
+// ---------------------------
+
+struct SpyElement : public ICanvasElement {
```
karensuzue: I think making these test helpers is a great idea!

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -0,0 +1,760 @@
...
+    // ID should not change even if we mount/unmount
+    REQUIRE_NOTHROW(layout.Unmount());
+    REQUIRE(layout.Id() == "stable-id-test");
+}
+
+TEST_CASE("WebLayout mountToLayout and unmount are safe", "[weblayout][dom]") {
+    WebLayout parent("parent");
+    WebLayout child("child");
+
+    REQUIRE_NOTHROW(child.MountToLayout(parent, Alignment::Center));
```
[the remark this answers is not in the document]
karensuzue: -R -C

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -0,0 +1,760 @@
...
+    REQUIRE_NOTHROW(child.Unmount());
+    REQUIRE_NOTHROW(child.Unmount());  // Idempotent unmount
+}
+
+TEST_CASE("WebLayout syncFromModel calls Apply", "[weblayout][dom]") {
+    WebLayout layout("sync-test");
+    // syncFromModel calls Apply internally; should not throw
+    REQUIRE_NOTHROW(layout.SyncFromModel());
```
karensuzue: Again, `REQUIRE_NOTHROW` doesn't test anything here.

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -0,0 +1,760 @@
...
+    REQUIRE(elem.mountCount == 0);
+}
+
+TEST_CASE("WebLayout can add and remove single element", "[weblayout][children]") {
+    WebLayout layout("add-remove-test");
+    MockDomElement elem("elem-1");
+
+    // Add element (note: in real scenario would need DOM to exist)
+    // The test checks internal state through Apply behavior
+    REQUIRE_NOTHROW(elem.MountToLayout(layout, Alignment::Center));
```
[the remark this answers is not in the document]
karensuzue: -R -C

tests/Interfaces/WebUI/WebLayoutTest.cpp
karensuzue: Overall, my main critique of these tests is the misuse of `REQUIRE_NOTHROW` when all of the functions are marked `noexcept`, and the fact that the tests only check whether the function runs without error rather than validating the behavior itself. There's also a lack of DOM‑level verification, which is especially important for a class like this that serves as the backbone for the other WebGUI classes.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -0,0 +1,386 @@
...
+  if (id.empty()) return false;
+
+  val el = mDocument.call<val>("getElementById", id);
+  if (el.isNull() || el.isUndefined()) {
+    return false;
+  }
+
+  // attach now; Apply() will re-check parent when re-applying
+  // if element already exists in this layout, this will just move it to the end
+  mRoot.call<void>("appendChild", el);
```
karensuzue: I assume this call doesn't create duplicates if 'el' already exists…

tests/Interfaces/WebUI/WebTextboxTest.cpp
```diff
@@ -0,0 +1,855 @@
...
+
+// ========================================================
+// Test 1: Constructor — defaults set correctly
+// ========================================================
+TEST_CASE("Constructor sets defaults correctly", "[webtextbox]")
+{
+    WebTextbox tb("hello");
+
+    CHECK(tb.GetText() == "hello", "initial text mismatch");
+    CHECK(tb.IsVisible() == true, "default should be visible");
```
karensuzue: If it hasn't been mounted yet, it would make more sense for me to have it start out invisible, but that's just me

tests/Interfaces/WebUI/WebButtonTest.cpp
```diff
@@ -0,0 +1,294 @@
...
+TEST_CASE("SetSize with zero dimensions", "[WebButton]") {
+  WebButton btn("test");
+  btn.SetSize(0, 0);
+  CHECK_MSG(btn.GetWidth() == 0, "width should be 0");
+  CHECK_MSG(btn.GetHeight() == 0, "height should be 0");
+}
+
+TEST_CASE("SetBackgroundColor does not crash", "[WebButton]") {
+  WebButton btn("test");
+  REQUIRE_NOTHROW(btn.SetBackgroundColor("#4CAF50"));
```
karensuzue: Same issue here, `SetBackgroundColor` doesn't throw exceptions. I'd check the styling at the DOM level instead

tests/Interfaces/WebUI/WebButtonTest.cpp
karensuzue: Some of your test cases misuse `REQUIRE_NOTHROW`, since `WebButton`'s member functions don't throw exceptions at all. I'd also suggest digging into the DOM to verify behavior, rather than only checking that a function runs without error (or that things look fine at the C++ level).

tests/Interfaces/WebUI/WebTextboxTest.cpp
```diff
@@ -0,0 +1,855 @@
...
+
+    tb.SetBackgroundColor("#AABBCC");
+    tb.ClearBackgroundColor();
+    tb.SetBackgroundColor("transparent");
+}
+
+// ========================================================
+// Test 16: Font fallback
+// ========================================================
+TEST_CASE("SetFallbackFontFamily does not crash", "[webtextbox][style]")
```
karensuzue: To be safe you shouldn't just make sure the function doesn't crash, you should also verify it at the DOM level.

tests/Interfaces/WebUI/WebTextboxTest.cpp
```diff
@@ -0,0 +1,855 @@
...
+
+    tb.SetMaxWidth(200.0f);
+    tb.SetWrap(true);
+    tb.SetWrap(false);
+}
+
+// ========================================================
+// Test 18: MountToLayout creates DOM element
+// ========================================================
+TEST_CASE("MountToLayout makes element appear in DOM",
```
karensuzue: This is a much more meaningful test case!

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,451 @@
...
+  int id = sNextRegistryId++;
+  sImageRegistry[id] = img;
+  return id;
+}
+
+void UnregisterImage(int id) {
+  sImageRegistry.erase(id);
+}
+
+void UpdateRegistryEntry(int id, cse498::WebImage* img) {
```
karensuzue: Something like this is useful for helping ensure that the registry remains free of dangling pointers :)

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,451 @@
...
+void WebImage::RemoveFromDom() {
+  if (mElement.isNull()) return;
+  val parent = mElement["parentNode"];
+  if (!parent.isNull() && !parent.isUndefined()) {
+    parent.call<void>("removeChild", mElement);
+  }
+}
+
+/// Removes element from DOM, deregisters from the event registry, and nulls the element.
+void WebImage::CleanupElement() {
```
karensuzue: Good idea!

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,451 @@
...
+      mElement(val::null()) {
+  mId = "webimage-" + std::to_string(mNextIdCounter++);
+  mRegistryId = RegisterImage(this);
+
+  val doc = GetDocument();
+  mElement = doc.call<val>("createElement", std::string("img"));
+  mElement.set("id", mId);
+  mElement.set("src", mSrc);
+  mElement.set("alt", mAltText);
+  doc["body"].call<void>("appendChild", mElement);
```
karensuzue: I would avoid immediately appending the image to the body - that's something that should be left to the user, and they'll likely instantiate a WebLayout first before mounting the image.

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+// ----- Sizing -----
+
+/// Sets the display size in pixels and applies object-fit: fill.
+void WebImage::SetSize(int width_px, int height_px) {
+  assert(width_px >= 0 && "SetSize: width must be non-negative");
+  assert(height_px >= 0 && "SetSize: height must be non-negative");
+  mWidth = width_px;
+  mHeight = height_px;
+  if (!mElement.isNull()) {
```
karensuzue: I would recommend outputting a message here for easier debugging in case mElement is null.

tests/Interfaces/WebUI/WebImageTest.cpp
```diff
@@ -0,0 +1,642 @@
...
+  REQUIRE_MSG(img.GetAltText() == "A cat",
+        "alt text mismatch");
+  CHECK_MSG(img.GetWidth() == 0,  "default width should be 0");
+  CHECK_MSG(img.GetHeight() == 0, "default height should be 0");
+  CHECK_MSG(img.GetOpacity() == 1.0, "default opacity should be 1.0");
+  CHECK_MSG(img.IsVisible() == true, "default should be visible");
+  CHECK_MSG(img.IsLoaded() == false, "default should not be loaded");
+  CHECK_MSG(img.HasError() == false, "default should have no error");
+  CHECK_MSG(!img.Id().empty(), "id should not be empty");
+  PASS();
```
karensuzue: PASS() does not seem necessary here.

tests/Interfaces/WebUI/WebImageTest.cpp
```diff
@@ -0,0 +1,642 @@
...
+  img.HandleError();
+  CHECK_MSG(img.HasError() == true, "should have error with BlankRect");
+  PASS();
+}
+
+// ========================================================
+// Test 23: SetPlaceholderColor (no crash, value stored)
+// ========================================================
+void test_set_placeholder_color() {
+  TEST("SetPlaceholderColor does not crash");
```
karensuzue: Similar to what I mentioned above, I would verify it myself to see whether anything changes at the DOM level.

tests/Interfaces/WebUI/WebImageTest.cpp
karensuzue: I would like to see a few more test cases specifically focused on the registry, since it is a crucial part of the class. For example, you could test whether a load/error event (which acts on registryId) affects the new/moved-to image, and that the moved-from/original object does not receive the event. It would also be good to verify that the destructor unregisters the image correctly, especially after ownership has been transferred from another image.

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    WebCanvas a("move-a");
+    a.AddElement(std::make_unique<LifetimeElement>());
+
+    // Move-construct should not throw.
+    REQUIRE_NOTHROW(WebCanvas{std::move(a)});
+
+    // Moved-from object should remain safe to use (even if it becomes "empty").
+    // We don't assert specific values, only that APIs remain valid.
+    REQUIRE_NOTHROW(a.ClearElements());
```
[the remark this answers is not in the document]
karensuzue: I think using REQUIRE_NOTHROW() is understandable when dealing with the external JS functions, since there's a possibility they might throw something, but I'm not sure whether they would throw a standard C++ exception that Catch can actually handle.

PR #38, opened 2026-02-17
Group1 Initial classes 

LimEddie19:
Includes the 5 classes: AnnotatedSet, TagManager, RobinHoodMaps, FeatureVector, and MemoryFactory. Along with 5 test classes for each of them


Luke-Antone (commented):
Great job on this code

tests/tools/Test_FeatureVector.cpp
AneeshJ10: Feedback

Testing and Error Checking:
You did a good job of providing tests for your member functions while checking for standard functionality.

Suggestions:
It might be a good idea to incorporate the size function and the main constructor into your tests. 
You could also include an edge case of creating a feature vector with an empty input.

source/tools/FeatureVector.hpp
```diff
@@ -0,0 +1,29 @@
...
+    public:
+        explicit FeatureVector(std::vector <double> values);
+
+        FeatureVector(const double* data, std::size_t size); //Main constructor
+
+        std::size_t size() const noexcept {return values_.size();}
+
+        double at(std::size_t index) const; //Access
+
+        double dot(const FeatureVector& other) const;
```
ZodiacalBee: add [[nodiscard]]

source/tools/FeatureVector.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+        void normalize();
```
ZodiacalBee: add comment mentioning that this is an inplace operation (though void implies that just make explicit documentation for all of this

source/tools/TagManager.cpp
ncharters: **Class Design**
- Identifiers:
  - Overall good function and variable identifiers, using t and s as variable names (Rows 96-97) would normally be considered not descriptive or too short but I think with the context it's probably not too bad, just something to consider.
- Magic number usage:
  - No magic numbers used.
- Interface:
  - No member variables are made public, and no functions are unnecessarily made public either. Function use is intuitive as well, good work.
- Comments:
  - I think some more comprehensive commenting would be helpful for those who are less familiar with the code, maybe a brief comment above each function in the .cpp file.
- Style:
  - All functions and member variables are named in a consistent style, and brackets are consistently placed.
- Appearance:
  - The code is very easy to look through with no inconsistent whitespace or indentation.

**Implementation**
- Compilation warnings:
  - No warnings are produced when compiling TagManager.cpp with the -Wall and -Wextra flags.
- Namespace usage:
  - All the code is correctly contained in the cse498 namespace.
- C++ standard:
  - In Row 16, your return statement for FindSet has a chance to return a nullptr, which I believe may cause undefined behavior when you then call find() on that nullptr in a different function.
- Modularity:
  - Good use of helper functions and avoiding any redundancy within your code.
- C++ constructs:
  - Usage of range based for loops is good. To touch back on the previous point of potentially calling find() on a nullptr, it may be worth looking into std::expected for use in FindSet, which would allow you to only return an ObjectSet if it wouldn't be null.
- STL algorithms:
  - I don't believe there are any standard library algorithms that would help improve the functionality of the current code.
- Pointers:
  - Currently no use of smart pointers, but I don't think that they are necessary when returning iterators from find() 
- Const correctness:
  - Make sure that parameters that don't change within a function are marked as const, for example if you're passing in a tag to a function and that tag will be the same after the function call, it can be marked const.

**Testing & Error Checking**
- Unit tests:
  - Very good work testing all of the public member functions, the one thing that I think would be helpful is testing a scenario in which AssertValidTag functions properly, such as calling OnTagAdded with an empty tag and making sure it's caught.
- Edge cases:
  - Your testing suite seems fairly comprehensive which is good, the one thing I could come up with that might be reasonable to test would be with multiple include and exclude tags at the same time, just to make sure it's handling that many query parameters properly.
- Asserts:
  - You have an assert in Row 10 to check for empty tags, but if a tag comes through that is empty, should it abort the program, or should it just send an error back to the user and not proceed with the code following that function call?
- Error handling:
  - You do a good job of using your AssertValidTag function to make sure that any tags being passed through your functions are valid and also use return statements to break out of a function where applicable.

source/tools/TagManager.cpp
```diff
@@ -0,0 +1,137 @@
...
+            const auto* s = FindSet(t);
+            if (!s)
+            {
+                return {}; // missing required tag
+            }
+            if (!baseSet || s->size() < baseSet->size())
+            {
+                baseSet = s;
+                baseTag = t;
+            }
```
ncharters: Potentially think about renaming the above variables s and t to be a little bit more descriptive to what they are.

source/tools/TagManager.cpp
```diff
@@ -0,0 +1,137 @@
+#include "TagManager.hpp"
+
+#include <cassert>
+#include <utility>
+
+namespace cse498
+{
+    void TagManager::AssertValidTag(std::string_view tag)
```
ncharters: Is this function ever tested properly in your test suite? Considering it is your main line of defense in other functions against invalid tag inputs it would probably be good to just make sure it works as intended.

ncharters (commented):
Overall, really good work on this class, just a few minor notes for potential adds and adjustments, but looks great.

source/tools/RobinHoodMap.cpp
```diff
@@ -0,0 +1,22 @@
...
+ * RobinHoodMap is a template class, so the full implementation lives in the
+ * header.  This file exists to verify that the template instantiates correctly
+ * at compile time.
+ * 
+ * 
+ * not completely sure if this is needed, but ill keep it for now. 
+ **/
+
+ 
+
```
murra381: Given that it is a template class, This is a perfectly fine cpp file and I like the addition of a static assert to verify compilation

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
...
+    /// @brief Find the value associated with a key.
+    /// @return Pointer to the value if found, nullptr otherwise.
+    [[nodiscard]] Value * Find(const Key & key) {
+      size_type idx = FindIndex(key);
+      if (idx == capacity_) return nullptr;
+      return &buckets_[idx].kv.second;
+    }
+
+    /// @brief Find the value associated with a key (const).
+    /// @return Pointer to the value if found, nullptr otherwise.
```
murra381: Your use of STL container conventions was consistent throughout. I would recommend changing your Insert, Find, Erase functions to also be lowercase to match the normal std::unordered_map conventions as well. Good job!

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+    /// @brief Remove all elements.  Bucket count stays the same.
+    void Clear() {
+      buckets_.assign(capacity_, Bucket{});
+      size_ = 0;
+    }
+
+    //  Iteration  (lowercase names required for range-based for loops)
+
+    iterator begin() {
```
murra381: Is there any downside to adding [[nodiscard]] here as well? I would add it.

tests/tools/RobinHoodMap.cpp
```diff
@@ -0,0 +1,867 @@
+// NOTE: CATCH_CONFIG_MAIN is defined in tests/core/WorldPosition.cpp;
+//       do NOT define it again here.
+
+// Generated some test cases with the use of an LLM to help ensure better test coverage *used opus 4.6 as well!
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+
```
murra381: I encountered no compilation errors, warnings, or exceptions. The code ran without issue. The test case file itself is well documented, thorough, and utilizes Catch2 test cases. It also utilizes CHECK and CHECK_FALSE to make sure the test cases still continue if one is proven to be incorrect. This is a well created test file that even checks for edge cases.

tests/tools/RobinHoodMap.cpp
```diff
@@ -0,0 +1,867 @@
...
+#include "../../source/tools/RobinHoodMap.hpp"
+
+#include <string>
+#include <vector>
+#include <set>
```
murra381: Unit Testing: All public methods you created were covered. (Insert, Find, Erase, etc)

Edge Cases: Satisfied with overall coverage (stress testing, negative keys, etc)

Assertions: I saw no issue with the assertions. They were used appropriately to guard against invalid arguments

Error Handling: I thought there was good coverage here too. I think you might be able to use std::expected to do some error handling in Rehash (if there is a bad allocation consider returning (std::unexpected(std::errc::not_enough_memory)) perhaps

murra381 (commented):
Great job on this. Code is well put together

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
...
+    };
+
+    //  Constants and member data
+
+    static constexpr size_type DEFAULT_CAPACITY = 16;
+    static constexpr double DEFAULT_MAX_LOAD_FACTOR = 0.60; // honestly not sure what the best load factor is, 
+                                                            // but based on my research, 0.60 is a good balance
+                                                            // between memory usage and performance.
+                                                            // this will be changed as an input parameter for anyone using the map.
+                                                            // so feel free to change it, but keep in mind that the default is 0.60.
```
Cobathon: Maybe add the source/research of how you came to this conclusion, for anyone looking to understand the unsureness that was mentioned.

Cobathon (commented):
This code is overall well written! Style formatting is consistent and it is very well documented. Small things here and there I commented on but nothing code breaking.

source/tools/AnnotationSet.hpp
```diff
@@ -0,0 +1,65 @@
+// AnnotationSet.h
```
tmaks2: Slight file name mismatch. Comment says .h, so you might want to rename it to .hpp for consistency.

source/tools/AnnotationSet.hpp
```diff
@@ -0,0 +1,109 @@
...
+#pragma once
+
+#include <cassert>
+#include <cstdint>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+#include <utility>
+#include <vector>
```
Ebizi: Class design:

1. informative identifiers - the identifiers are good, there isn't anything that seems vague or unclear. you could maybe change NotifyAdded and NotifyRemove to specify NotifyTagAdded and NotifyTagRemoved, but I think in the context of this being in the annotationset class it would be fine as is

2. avoids magic numbers - this doesn't have any numbers so you pass on this

3. intuitive and functional interface - it clears out everything that's mentioned, so member variables are private and the functionality is obvious. good work

4. informative comments - the comments are good. they quickly explain what's going on without holding your hand on what's happening

5. consistent style - styling does seem consistent, I think it's pretty consistently using PascalCase

6. visual appearance - i personally scanned through the code fairly well

Implementation:

1. no warnings during compilation - it compiled fine on my end so this should be good

2. proper uses of namespaces - seems correct in that everything is in cse498 and avoids the using namespace std

3. adheres to c++ standard - overall everything looked fine, congrats

4. appropriate modularity - this looks fine as well, you separated functions into their own distinct roles like AssertValidTag and NotifyAdded + NotifyRemoved instead of pooling them all into one big megafunction

5. uses standard library algorithms - i did not see any algorithm usage so you pass on this

6. pointer handling - no manual new or delete noticed, so this seems fine

7. const correctness - this is handled fine, things like Size and whatnot are marked const when needed

8. uses modern c++ constructs - this also seems fine, nothing stands out as egregious and you seem to use things like std::string_view correctly

testing:

1. unit testing - unit tests seem good, they cover all the basics like construction, duplicates, clearing, etc

2. edge cases - i could not notice any major edge case that was not tested, the large element one tested the code as it scales fine

3. assertions - nice assertions, these seemed fine, congrats!

4. error handling - seemed fine

overall - congrats, this is really well done

source/tools/AnnotationSet.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+// Notify TagManager that a tag was removed from this object
+void cse498::AnnotationSet::NotifyRemoved(std::string_view tag) noexcept {
+    // if (mManager) {
+
+    //     mManager->OnTagRemoved(mOwner, tag);
+    // }
+}
+
+void cse498::AnnotationSet::AssertValidTag(std::string_view tag) {
```
tmaks2: Indentation and spacing appears off with this. Looks like there is an extra tab on the assert line, end bracket spacing doesn't match the rest of your functions, and the blank line above the assert is kind of awkward.

source/tools/AnnotationSet.hpp
```diff
@@ -0,0 +1,65 @@
...
+            for (const auto& t : mTags) out.push_back(t);
+            return out;
+        }
+
+        using const_iterator = std::unordered_set<std::string>::const_iterator;
+        const_iterator begin() const noexcept { return mTags.begin(); }
+        const_iterator end() const noexcept { return mTags.end(); }
+
+    private:
+        ObjectId mOwner{};
```
tmaks2: I think this should be up with the other member variables rather than separated.

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
+/**
+ * This file is part of the Fall 2026, CSE 498, section 2, course project.
+ * @brief A high-performance hash map using open addressing with robin hood hashing.
```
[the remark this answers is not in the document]
aryaplayzz: okay will do! thanks

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
...
+ * @note Status: PROPOSAL
+ * @citation: Used Claude Code Opus 4.6 to help me understand the algorithm and implement it.
+ * @note: Took around 10 total hours to implement this. (Including test cases and documentation)
+ **/
```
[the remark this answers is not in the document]
aryaplayzz: yep will do!

source/tools/RobinHoodMap.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+    static constexpr size_type DEFAULT_CAPACITY = 16;
+    static constexpr double DEFAULT_MAX_LOAD_FACTOR = 0.60; // honestly not sure what the best load factor is, 
+                                                            // but based on my research, 0.60 is a good balance
+                                                            // between memory usage and performance.
+                                                            // this will be changed as an input parameter for anyone using the map.
+                                                            // so feel free to change it, but keep in mind that the default is 0.60.
+
+    std::vector<Bucket> buckets_;  
+    size_type size_ = 0;
```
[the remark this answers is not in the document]
aryaplayzz: okay thank you!

Luke-Antone (commented):
Great job overall, my commets are gernaly small tweeks as the code is well made

source/tools/MemoryFactory.hpp
finfroc3: Overall, a very good class! It took me a second to wrap my head around how it works, but the way it was written didn't impede my understanding at all. The naming conventions you used are consistent, the names themselves are simple and descriptive, the comments are placed where they are necessary and aren't redundant, and the overall formatting is very clean. I especially like some of the safeguards you put in place. The deletion of the copy constructor and the asserts make it much harder for the user to misuse the class and assure the class's specifications are correct before anything begins running. The use of smart pointers enforces what we learned in class, and the custom deleter properly cleans up everything and prevents memory leaks. I don't want to write a review of exclusively praise if I can help it, but looking through, I really can't see much wrong with it. Maybe go into a little more detail on some of the comments?

source/tools/TagManager.cpp
```diff
@@ -0,0 +1,137 @@
...
+                return {}; // missing required tag
+            }
+            if (!baseSet || s->size() < baseSet->size())
+            {
+                baseSet = s;
+                baseTag = t;
+            }
+        }
+
+        out.reserve(baseSet->size());
```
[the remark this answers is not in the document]
LimEddie19: Added an assert before the baseSet to check for null

source/tools/TagManager.cpp
LimEddie19: I really liked this review. It gave me a good sense of what I was missing overall. It's incredibly detailed and show me that you understood. Which also tells me your suggestions have more weight to them. Great work

source/image_manager_test
nahumj: Remove this file.

tests/tools/MemoryFactory.cpp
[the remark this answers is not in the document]
nahumj: Very nice class!

tests/tools/TagManagerTest.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+TEST_CASE("TagManager: OnTagAdded makes HasTag true and Query returns owner", "[TagManager]")
+{
+    TagManager tm;
+    TagManager::ObjectId a = 1;
+
+    tm.OnTagAdded(a, "red");
+
+    REQUIRE(tm.HasTag(a, "red") == true);
```
nahumj: You don't need to check if a bool is equal to true.

source/tools/RobinHoodMap.hpp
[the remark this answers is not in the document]
nahumj: All use of AI should be cited with a link to the conversation!

nahumj (approved):
Please see discord for grade.

Ebizi (commented):
Hi, apologies for the late review as I had set this to pending instead of submitting it. My bad, and thank you for your patience!

source/tools/TagManager.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+        /// Errors that may be returned by FindSet().
+        enum class FindSetError { EmptyTag, TagNotFound };
+
+        TagManager() = default;
+
+        void OnTagAdded(ObjectId owner, std::string_view tag);
+        void OnTagRemoved(ObjectId owner, std::string_view tag);
+        
+        std::vector<ObjectId> Query(
```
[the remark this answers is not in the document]
LimEddie19: Thank you for the suggestion. I felt that using a set would create unnecessary construction and hashing overhead on the caller over a vector since we are querying through small sets of tags. Of course, if there is something I am missing, feel free to correct me on this.

source/tools/TagManager.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * Finds the objects associated with a/multiple tags.
+     *
+     * Performs a lookup in the internal tag-to-object map.
+     *
+     * @param tag The tag to search for.
+     * @return Pointer to the ObjectSet associated with the tag, or nullptr if the tag does not exist.
+     */
+    std::expected<const TagManager::ObjectSet*, TagManager::FindSetError>
```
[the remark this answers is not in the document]
LimEddie19: Created a reference wrapper around TagManager::ObjectSet to return reference instead of copy

source/tools/TagManager.hpp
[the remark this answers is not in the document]
LimEddie19: I changed it so that all throws instead have to return an std::expected

tests/tools/TagManagerTest.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+TEST_CASE("TagManager: OnTagRemoved removes membership for that tag", "[TagManager]")
+{
+    TagManager tm;
+    TagManager::ObjectId a = 1;
+
+    tm.OnTagAdded(a, "red");
+    tm.OnTagRemoved(a, "red");
+
+    REQUIRE(tm.HasTag(a, "red") == false);
```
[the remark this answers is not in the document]
LimEddie19: Fixed all instances to REQUIRE(condition == false) to REQUIRE_FALSE(condition)

tests/tools/TagManagerTest.cpp
LimEddie19: Changed all instances of checking bool = true to just REQUIRE(condition).

source/tools/TagManager.cpp
```diff
@@ -0,0 +1,255 @@
...
+     *
+     * If includeTags is empty, all known objects are considered as the base set.
+     *
+     * @param includeTags Tags that objects must contain.
+     * @param excludeTags Tags that objects must not contain.
+     * @return A vector of object IDs matching the query conditions.
+     *
+     * @throws std::invalid_argument if any tag is empty.
+     */
+    std::vector<TagManager::ObjectId> TagManager::Query(
```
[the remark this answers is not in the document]
LimEddie19: I split the Query function up into multiple helper functions with the main Query function looking cleaner and easier to understand

tests/tools/TagManagerTest.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <stdexcept>
+#include <string_view>
+#include <unordered_set>
+#include <vector>
+
+using cse498::TagManager;
+
+// Helpers Functions
+static std::unordered_set<TagManager::ObjectId>
+ToSet(const std::vector<TagManager::ObjectId>& v)
```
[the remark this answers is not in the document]
LimEddie19: Changed ToSet to reflect vectors instead of sets

PR #39, opened 2026-02-17
Group14

lunncurt:
# Group 14 Base Classes
###     -  Test Makefile updated to support new third-party content, may need to be reviewed.


group_specific_content/Group-14/Group14InitialClasses/EventQueue.hpp
```diff
@@ -0,0 +1,81 @@
+/**
+ * EventQueue: store events and always give you the next event to run.
+ */
+
```
ZhixiangMiao: First of all, great job on passing all test cases in the EventQueue class. However, you should add more comments to describe what each function does exactly (for example, explaining what the code in your function contributes to what you want your function do) to help other people understand better.

group_specific_content/Group-14/Group14InitialClasses/EventQueue.hpp
```diff
@@ -0,0 +1,81 @@
...
+  class EventQueue {
+  private:
+    std::priority_queue<Event> events;
+    std::unordered_set<EventID> cancelled;
+    EventID next_id = 0;
+    std::uint64_t next_order = 0;
+
+    // remove cancelled events at the top
+    void CleanTop() {
```
ZhixiangMiao: In your CleanTop() function, you didn't fully cleaned the cancelled event, although you popped the cancelled events at the top using events.pop(), and in your code "cancelled.count(events.top().id) > 0" what I interpreted this code is to find the event id in your cancelled set, if it is inside your cancelled set then you pop() the event from events priority queue (since it is cancelled event id, so you want this event to be removed). But also, since you successfully cancelled the event and popped it from the events priority queue, you should also remove the id in your cancelled set. Since the event is already cancelled and removed, you can release that id and later you can store other event ids you want to cancel inside your cancelled set.

source/tools/InteractiveWorld/DataMap.hpp
GumdropHorse: Hi! Here are my peer review notes on the DataMap Class.
Below I have the full rubric, but just to help separate it a bit, here is the one change I would suggest making:
**Class Design**
Consistent style.
Overall, style choices are very consistent. All variable/class names are one word, functions start with Caps and variables are fully lower case. The one suggestion I would make is to change the member variable map to be m_map to differentiate it from normal variables.

Overall it is a well written class! Good work!


Full Rubric Review:

**Class Design**
Informative identifiers.
Identifiers are clear and helpful.

Avoids “magic numbers”.
There are no magic numbers.

Intuitive and functional interface.
Class only exposes necessary functionality, is easy to use correctly and hard to misuse. Member variables are private. There are obvious functions to use for common tasks, however there are no default arguments or parameters. However, this might be okay as the functions aren’t really set up in a way where it makes sense to have default arguments. None of the functions are misleading.

Informative comments.
The comments helped me to understand the code rapidly, and it properly avoided redundant and obvious comments.

Consistent style.
Overall, style choices are very consistent. All variable/class names are one word, functions start with Caps and variables are fully lower case. The one suggestion I would make is to change the member variable map to be m_map to differentiate it from normal variables.

Visual appearance.
I would suggest adding newline separation between all of the classes, and maybe double new line separation between the private/public sections, and increase to double/triple newline separation between the struct and the class. This would just help break things up a bit more and help with readability. Overall though it looks pretty good.


**Implementation**
No warnings during compilation.
I did not get any errors or warning when I compiled the code.

Proper uses of namespaces.
All code is within namespace cse498, and avoids using namespace std.

Adheres to the C++ standard.
The code avoids undefined behavior, and compiles and runs on my compiler and machine.

Appropriate Modularity.
Common procedures are grouped into useful helper functions, there is a helper struct used where needed and unnecessary code duplication is avoided.

Uses modern C++ constructs.
The code adheres to proper use of c++ constructs.

Uses standard library algorithms.
Not applicable, the code has no use for standard library algorithms.

Pointer handling.
There are pointers used, but they are used properly and carefully.

const correctness.
Const correctness is maintained throughout the code.


**Testing and Error Checking**
Unit testing.
There is not a specific test for the constructor, but it is just the default constructor and is handled by all of the other tests. There are full tests for all of the other public member functions, and they are well organized. They also trigger cases that exercise all of the private portions of the code.

Edge cases.
All edge cases seem to be covered by the tests.

Assertions.
Assert statements are used throughout the class, particularly when it comes to the templated functions.

Error handling.
Error cases are properly handled with asserts.

group_specific_content/Group-14/Group14TestCases/EventQueue.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+TEST_CASE("PeekNextEvent returns smallest time") {
+    EventQueue q;
+
+    q.ScheduleEvent(10, [](){});
+    q.ScheduleEvent(5, [](){});
+
+    CHECK(q.PeekNextEvent().time == 5);
+
```
ZhixiangMiao: Test looking great, however, you should also check and peek the time of the event if it's equal to 10, since you scheduled an event at time 10. To check and test everything you scheduled is a good practice.

tests/InteractiveWorld/EventQueue.cpp
```diff
@@ -0,0 +1,72 @@
...
+TEST_CASE("PeekNextEvent returns smallest time") {
+    EventQueue q;
+
+    q.ScheduleEvent(10, [](){});
+    q.ScheduleEvent(5, [](){});
+
+    CHECK(q.PeekNextEvent().time == 5);
+
+    // After peek event, it should still exist
+    CHECK(q.Empty() == false);
```
ZhixiangMiao: Test looking great, however, you should also check and peek the time of the event if it's equal to 10, since you scheduled an event at time 10. To check and test everything you scheduled is a good practice.

group_specific_content/Group-14/Group14InitialClasses/ExpressionParser.hpp
```diff
@@ -0,0 +1,108 @@
...
+//  and returns a function object that takes a
+//  std::unordered_map<string, double>
+//  as input and evaluates the given expression
+
+namespace cse498 {
+
+  class ExpressionParser {
+  private:
+    // Collect unique variable identifiers referenced by the expression string.
+    std::vector<std::string> CollectVariables(const std::string& input_expression) {
```
AbbyFranz: Const correctness - This member function can be made const because it does not change the state of the class.

source/tools/InteractiveWorld/ExpressionParser.hpp
```diff
@@ -0,0 +1,108 @@
...
+//  and returns a function object that takes a
+//  std::unordered_map<string, double>
+//  as input and evaluates the given expression
+
+namespace cse498 {
+
+  class ExpressionParser {
+  private:
+    // Collect unique variable identifiers referenced by the expression string.
+    std::vector<std::string> CollectVariables(const std::string& input_expression) {
```
AbbyFranz: Const correctness - This member function can be made const because it does not change the state of the class.

source/tools/InteractiveWorld/ExpressionParser.hpp
AbbyFranz: Appropriate Modularity - Parts of the Parser function can be split up into several helper functions. Although nothing is duplicated, I think it would help with readability to split this function up into smaller ones. For example within the returned lambda funciton, you could build your bindings in a seperate function and evaluate the expression in a seperate function. This would help shink your lambda down into only a couple of lines and keep it short.

source/tools/InteractiveWorld/EventQueue.hpp
```diff
@@ -0,0 +1,80 @@
...
+    bool operator<(const Event& other) const {
+      if (time != other.time) return time > other.time;  // smaller time first
+      return order > other.order;
+    }
+  };
+
+  class EventQueue {
+  private:
+    std::priority_queue<Event> events;
+    std::unordered_set<EventID> cancelled;
```
ZhixiangMiao: Informative identifiers: 
Using the name like: cancelled_ids, cancelled_id_set, etc. would be better for understanding.

tests/InteractiveWorld/EventQueue.cpp
```diff
@@ -0,0 +1,72 @@
...
+    q.ScheduleEvent(5, [](){});
+    q.ScheduleEvent(7, [](){});
+
+    CHECK(q.PopNextEvent().time == 5);
+    CHECK(q.PopNextEvent().time == 7);
+    CHECK(q.PopNextEvent().time == 10);
+    CHECK(q.Empty() == true);
+}
+
+TEST_CASE("CancelEvent skips cancelled event and returns next event (id2)") {
```
ZhixiangMiao: Edge case && Error handling: 
I suggest you try to cancel an non-existed event id for a test case. And I suggest that when you cancel an non-existed event id, an error should be thrown.

source/tools/InteractiveWorld/Scheduler.hpp
```diff
@@ -0,0 +1,123 @@
...
+            );
+        }
+
+        std::vector<Process> processes; // List of processes with their priorities
+        std::mt19937 rng{ std::random_device{}() }; // Random number generator for weighted random selection
+
+    public:
+        // Add a process with an associated priority
+        void AddProcess(ProcessID id, float priority) {
+            assert(priority > 0.0f && "Scheduler priority must be positive");
```
sitarakb: Avoids "magic numbers": I noticed a few times throughout the class that priority is compared to 0.0f. I don't personally view this as a magic number because the meaning of the number is obvious from context, but it still could be important to note.

source/tools/InteractiveWorld/Scheduler.hpp
```diff
@@ -0,0 +1,123 @@
+/**
+ * Scheduler: Schedules a set of processes based on a "priority" heuristic and returns the next process to run
```
sitarakb: Informative comments, consistent style, & visual appearance: As someone new to the code, I believe the comments were informative but not redundant or obvious. Additionally, the formatting choice were consistent (with the use of the camelCase naming convention), making the code easy to scan through, alongside the good use of whitespace, indentation, & line breaks.

source/tools/InteractiveWorld/Scheduler.hpp
```diff
@@ -0,0 +1,123 @@
...
+ */
+
+#pragma once
+
+#include <vector>
+#include <random>
+#include <cassert>
+#include <algorithm>
+
+namespace cse498 {
```
sitarakb: No warnings during compilation, proper use of namespaces, & adheres to the C++ standard: the code compiled without producing errors, all the code is within namespace cse498, 'using namespace std' is not used, and I cannot see any places where undefined behavior is not avoided! Good job!

tests/InteractiveWorld/Scheduler.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_CASE("Next returns a valid process ID") {
+    Scheduler s;
+
+    s.AddProcess(10, 1.0f);
+    s.AddProcess(20, 1.0f);
+    s.AddProcess(30, 1.0f);
+
+    for (int i = 0; i < 100; ++i) {
+        ProcessID id = s.Next();
```
sitarakb: Edge cases, assertions, & error handling: examining the test cases, it looks like every edge case I could personally think of is covered! Great job! Assertions are made properly throughout the code, & error handling is checked for in the class in AddProcess, Next, & UpdatePriority using asserts.

source/tools/InteractiveWorld/FunctionSet.hpp
joeyhyun1: I have reviewed this file and it looks well done for its purpose. Nothing jarring really stood out to me in terms of bad coding practice and the code all compiled without issues. All the test cases also passed so no issues there. Overall I can't really find anything that I would change as it all works and does its job well.

source/tools/InteractiveWorld/Scheduler.hpp
```diff
@@ -0,0 +1,123 @@
...
+                [id](const Process& p) { return p.id == id; }
+            );
+        }
+
+        std::vector<Process> processes; // List of processes with their priorities
+        std::mt19937 rng{ std::random_device{}() }; // Random number generator for weighted random selection
+
+    public:
+        // Add a process with an associated priority
+        void AddProcess(ProcessID id, float priority) {
```
paulbui081: Appropriate Modularity: Scheduler.hpp properly groups different procedures (AddProcess(...), Next(), UpdatePriority(...), etc.) into their own helper functions, allowing for the code to be re-usable and look visually clean.

source/tools/InteractiveWorld/Scheduler.hpp
```diff
@@ -0,0 +1,123 @@
...
+        // Const version of FindProcess for use in const methods
+        auto FindProcess(ProcessID id) const {
+            return std::find_if(
+                processes.begin(),
+                processes.end(),
+                [id](const Process& p) { return p.id == id; }
+            );
+        }
+
+        std::vector<Process> processes; // List of processes with their priorities
```
paulbui081: Intuitive and functional interface: Scheduler.hpp exposes only what's necessary for proper use, keeping internal information such as process and the number generator member variables private to prevent misuse.

group_specific_content/Group-14/Group14InitialClasses/FunctionSet.hpp
```diff
@@ -0,0 +1,82 @@
...
+                         [id](const Entry& e) { return e.id == id; }),
+          functions.end());
+      return functions.size() != old_size;
+    }
+
+    // Remove all functions and reset the ID.
+    void Clear() { 
+      functions.clear(); 
+      next_id = 0; 
+    }
```
marksandychez: I think this code is completely functionally correct; however, since the AddFunction returns an ID back to the caller, I think this code has the potential for a weird logic error if two different processes are using the function set.

For example, Process1 initializes the FunctionSet object and adds a function in at Index 0, then Process 2 clears the functions vector, and adds it's own function at index 0. In this case, depending on what the other process code looks like, Process 1 could think they had some function at index 0 and cause logic issues for the other process when accessing "it's" function. 

Considering how many IDs you can make with uint64, just removing the 
`next_id = 0` would ensure each function gets a unique id to reference back to regardless of how often it gets cleared by other processes. Technically this could overflow and would need guards, but it would be a massive amount of time before that id limit was reached.

marksandychez (commented):
Overall, the FunctionSet class is very well made and meets all major criteria.

- Class Design
The class uses informative identifiers, avoids Magic Numbers, correctly houses public and private functions/member variables, and uses a consistent style with good whitespace usage

- Implementation
There are no warnings during compilation, the namespace is correctly used, and makes very effective use of modern C++ features and algorithms. 

- Testing
The testing file makes a series of good logical checks to ensure that the class alters it's member variables as expected. There are uses of assertions and checks to validate this behavior. There are several more edge cases that could be addresses (inline comment on the file), but does a good job to test a wide range of inputs to the class.

Really well done!

group_specific_content/Group-14/Group14InitialClasses/DataMap.hpp
```diff
@@ -0,0 +1,73 @@
...
+		    if (!Has(key)) { return false; }
+		    map.erase(key);
+		    return true;
+	    }
+	    /// Clear out map of all data
+	    void Clear() { map.clear(); }
+	    /// Checks if the map is empty
+	    bool Empty() { return map.empty(); }
+	    /// Returns the number of elements in the map
+	    int Size() { return map.size(); }
```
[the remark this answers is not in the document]
kianamay: wrong return type, .size() returns size_t not int

group_specific_content/Group-14/Group14InitialClasses/DataMap.hpp
```diff
@@ -0,0 +1,73 @@
...
+		    // Return the value of the found pair
+		    assert((typeid(T) == it->second.type && "Type mismatch on Get!"));
+		    return *std::any_cast<T>(&it->second.value);
+	    }
+	    /// Check to see if key is in the map
+	    bool Has(const std::string& key) const { return map.find(key) != map.end(); }
+	    /// Remove key from map. Returns bool if successful
+	    bool Remove(const std::string& key) {
+		    // Make sure key exists
+		    if (!Has(key)) { return false; }
```
kianamay: the error handling in Remove is inconsistent with other functions like Get and Set which are using asserts I would aim for consistency if possible.

group_specific_content/Group-14/Group14InitialClasses/DataMap.hpp
```diff
@@ -0,0 +1,73 @@
+/**
+ * This file is for the Fall 2026 CSE 498 section 2 Capstone project.
+ * @brief Maps strings to arbitrary data types. Caller is not required to perform casts. Types in the DataMap are
+ * explicit at the call site. Any type mismatches are handled up front by DataMap.
+ **/
+
```
kianamay: Informative Identifiers: Yes names were clear and consistent in formatting
Magic Numbers: None present
Intuitive Interface: map is private and functions are obvious to use
Informative Comments: Comments are useful, not redundant
Style Consistency: Style is consistent but will need to follow new company A styling moving forward 
(Code Style: Clang Format
Variable name: camelCase (without m-prefix for member variables) TBD
Function name: PascalCase
Global Variables: FULL_UPPERCASE
Libraries: lowercase)
Visual appearence: Code is well structured and clear when scanning through

group_specific_content/Group-14/Group14TestCases/DataMap.cpp
```diff
@@ -0,0 +1,109 @@
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "../../group_specific_content/Group-14/Group14InitialClasses/DataMap.hpp"
+
+TEST_CASE("Test DataMap Set and Get Functions")
```
kianamay: Unit testing: Unit tests cover all public functions and are well-organized by function
Edge cases: Tests cover basic edge cases like empty maps, removing non-existent keys, and clearing empty maps, but missing tests for things like type mismatch scenarios, updating existing values with different types, and empty strings as keys
Assertions:  Asserts are used in most functions except Remove which returns false instead of asserting 
Error handling: Looks good but Remove() is inconsistent and may need to be switched to an assert

source/tools/InteractiveWorld/ExpressionParser.hpp
```diff
@@ -0,0 +1,123 @@
...
+    std::vector<std::string> vars;
+    std::unordered_set<std::string> seen_vars;
+    while (lexer.Any()) {
+      const auto &token = lexer.Use();
+      if (token.id != emplex::Lexer::ID_VARIABLE) {
+        continue;
+      }
+
+      // Keep first-seen order while removing duplicates.
+      if (seen_vars.insert(token.lexeme).second) {
```
nahumj: I would move de duplicating to s second loop (as it isn't connected to the tokenizing).

source/tools/InteractiveWorld/ExpressionParser.hpp
```diff
@@ -0,0 +1,123 @@
...
+        vars.push_back(token.lexeme);
+      }
+    }
+
+    return vars;
+  }
+
+public:
+  const std::function<double(const std::unordered_map<std::string, double> &)>
+  Parser(const std::string &input_expression) {
```
nahumj: This function is doing far to much and the fact that you have a monolithic function makes it hard to test the individual components.

tests/InteractiveWorld/Scheduler.cpp
```diff
@@ -0,0 +1,244 @@
...
+    CHECK(s.HasProcess(2) == false);
+}
+
+TEST_CASE("UpdatePriority does not create new processes") {
+    Scheduler s;
+
+    s.AddProcess(1, 1.0f);
+    CHECK(s.HasProcess(2) == false);
+
+    // This should assert in debug mode; in release it does nothing
```
nahumj: This comment is wrong

tests/InteractiveWorld/Scheduler.cpp
nahumj: Many of the test cases here are not respectful of the amount of processing time it takes to run some of these tests. Slow tests should be tagged so that they needn't be run every time. And functionality tests should be separated from stress tests.

PR #48, opened 2026-03-09
Remove generated SDL artifacts (`/third-party/SDL2/` and `/third-party/sdl-build/`)

Mehver:
SDL2 needs to be installed via the Makefile (`make install`). The related files currently kept in this repository were generated during installation and should not be committed/synchronized.

Changes:
- Remove `/third-party/SDL2/` and `/third-party/sdl-build/`
- Added these two path to `.gitignore`
- Added instructions in `make help` outputs for SDL2's install/uninstall

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,3 @@
-# SDL2 local installment files
-third-party/SDL2/
-third-party/sdl-build/
-
 # Prerequisites
 *.d
 


Mehver:
> I'm not sure when it got added there

I checked the file history, and it comes from a commit ([Commit `0279a8c`](https://github.com/CSE498/Spring2026-CompanyA/commit/0279a8ce9e602a8fb398ed084de7a2e8cd69c9dd)) that fixed some code during a peer review. A classmate likely submitted it unintentionally.

> but it hasn't caused any issues on my end. Since it was already there, I just skipped `make install` and ran the tests.

I think if SDL2 has already been installed on your system, this code wouldn’t cause any problems. However, it can produce errors on a fresh system based on what I got in a new GitHub Workspace.

My guess is that SDL2 isn’t completely self-contained; it needs to be adapted to different operating systems during installation, for example by setting up hooks or configuring environment variables. Also, the installation process involves a large amount of local compilation, which suggests that these files are platform-specific.

Mehver:
@nahumj You can try running `make test` again; it should pass now.

PR #57, opened 2026-03-17
Initial CMake version

lunncurt:
# If you encounter a build error

(first try to do a make install, as the sdl2 install is removed from this branch)

Please leave a comment with:

- Your system / environment (examples: macOS (ARM), macOS (Intel), WSL, Ubuntu, Windows)
- The exact command(s) you ran
- The compiler/toolchain being used, if known (`clang++ --version`, `g++ --version`, etc.)
- The full error output
- Whether the issue happened during configure, build, or test
- Any setup steps you took before the error occurred

Also helpful if applicable:

- Whether the main branch Make-based build worked on your machine
- Any local changes you had before testing
- A screenshot or pasted terminal output

This is my first time working with CMake on a project this large, so any help or suggestions are more than welcome.


karensuzue:
Aside from the issue reported by @sitarakb above (and in [#58](https://github.com/CSE498/Spring2026-CompanyA/issues/58)), everything seems to work fine for me! I'm using WSL Ubuntu. 

It would be great to have some centralized instructions on how to build with CMake and install all the required dependencies (like SDL and Emscripten). I'm guessing not everyone is familiar with CMake and I rarely use it myself :)

Here's what I did to get things running:
- Ran `cmake -S . -B build` (this generates a new `build` folder)
- Then `cd build` and ran `make` (or alternatively, `cmake --build build` from root).
- To run the `simple` executable, go to `/source`
- To run the tests, go to `/test/build` to find `./run_tests`. For the WebGUI tests you'll need to set up the server as usual.

The test cases for the GUI group (which previously worked) failed when I ran `./run_tests`. I'll open a new issue for that. 

Also, @sitarakb, I think you might still be running the old Makefiles?

sitarakb:
@karensuzue thank you for the instructions! I was in fact still running the old Makefiles but now I've got the cmake files running instead. I am also getting the GUI test fails. I'll respond on the issue you opened about it.

PR #68, opened 2026-03-25
Group15

GumdropHorse:
This is Group 15's pull request for the 3/25 due date. For review, we are submitting Random.hpp and WeightedSet.hpp from our original core C++ classes. 
Additionally, to have view a demo of our module, follow the CMake instructions to build the program, then cd into the demo folder, and run the generated Group15_demo executable.
The files that are apart of our module include:
   demo/ Group15_demo.cpp
   source/core/rooms/ (everything in this folder)
   source/core/items/ (everything in this folder)
   source/core/BSP-Dungeon.hpp
   source/core/DungeonBase.hpp
   source/core/RoomHolder.hpp
   source/core/WorldGeneration.hpp
   source/core/Worlds/DungeonOne.hpp
   tests/tools/GenerativeWorld/BSP-Dungeon-Seed-Info/ (everything in this folder)


source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
...
             // changes each time a number is generated
-            struct xoshiro256pp_state {
+            struct m_Xoshiro256ppState {
                 uint64_t s[STATE_NUMBER];
             };
 
-            struct splitmix64_state {
+            struct m_Splitmix64State {
                 uint64_t s;
             };
```
motowsk2: These internal helper structs are using names like:
- m_Xoshiro256ppState
- m_Splitmix64State
Those m_ are used for memeber variables, not type names. Is there a certain reason you named these this way?

source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
 
             // keeps an internal state of 4 64 unsigned 64-bit ints
             // changes each time a number is generated
-            struct xoshiro256pp_state {
+            struct m_Xoshiro256ppState {
                 uint64_t s[STATE_NUMBER];
```
motowsk2: I would suggest using an std::array instead of a raw array. Raw arrays lack bound safety and STL compatibility, compared to std::array which enables range-based loops, integrates with STL alogrithms, and improvs type safety and readability. My suggestion would be to change this to:
std::array<uint64_t, STATE_NUMBER> s{};

source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
...
+                struct m_Splitmix64State sm = {m_seed};
 
-                state.s[0] = splitmix64(sm);
-                state.s[1] = splitmix64(sm);
-                state.s[2] = splitmix64(sm);
-                state.s[3] = splitmix64(sm);
+                state.s[0] = m_Splitmix64(sm);
+                state.s[1] = m_Splitmix64(sm);
+                state.s[2] = m_Splitmix64(sm);
+                state.s[3] = m_Splitmix64(sm);
```
motowsk2: I would suggest replacing the repeated assignments with a loop. Manually assigning each index introduces unnecessary duplication and makes the code harder to maintain if the state size changes. Using a range-based loop is cleaner and more scalable. My suggestion would be to change this to:

for (auto &value : state.s) 
{
    value = m_Splitmix64(sm);
}

source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
...
             }
 
-            // Performs a left rotation on x by k bits
-            uint64_t rol64(uint64_t x, int k) {
+            /// @brief Performs a left rotation on x by k bits
+            /// @param x and k, x is the value being rotated and k is how much it is rotated by
+            /// @return A rotated x value
+            uint64_t m_Rol64(uint64_t x, int k) {
```
motowsk2: Same thing with the m_ again. Those m_ are used for memeber variables, is there a certain reason you named these this way?

source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
...
-                    throw std::runtime_error("cse498::Random::rol64(): k = 0 is undefined behavior");
+                    throw std::runtime_error("cse498::Random::m_Rol64(): k = 0 is undefined behavior");
                 }
                 else if (k == 64) {
-                    throw std::runtime_error("cse498::Random::rol64(): k = 64 is undefined behavior");
+                    throw std::runtime_error("cse498::Random::m_Rol64(): k = 64 is undefined behavior");
                 }
 
                 return (x<<k) | (x >> (NUM_BITS-k));
             }
```
motowsk2: I would suggest avoiding runtime exceptions in this low-level rotation function. Since this function is used internally and the shift values are controlled, throwing exceptions adds unnecessary overhead and complexity. In performance-critical code like RNGs, it’s better to rely on correct usage rather than runtime checks. My suggestion would be to simplify this to:

static constexpr uint64_t Rol64(uint64_t x, int k) noexcept 
{
    return (x << k) | (x >> (NUM_BITS - k));
}

source/tools/Random.hpp
```diff
@@ -49,53 +49,60 @@ namespace cse498 {
...
 
-            // Generates a random number
-            uint64_t xoshiro256pp(struct xoshiro256pp_state &state) {
+            /// @brief Generates a random number
+            /// @param the xoshiro state being used to generated the number
+            /// @return a randomly generated uint64_t value
+            uint64_t m_Xoshiro256pp(struct m_Xoshiro256ppState &state) {
                 uint64_t *s = state.s;
```
motowsk2: I would suggest avoiding the use of a raw pointer here. Using a reference is safer and it makes ownership/intent clearer. My suggestion would be to change this to:

auto &s = state.s;

source/tools/Random.hpp
```diff
@@ -107,135 +114,95 @@ namespace cse498 {
...
             }
 
-            struct xoshiro256pp_state m_rng;
+            struct m_Xoshiro256ppState m_rng;
             bool m_used = false;
 
-            void check_rng() {
+            /// @brief Checks if the rng has been initalized or not
+            void m_CheckRng() {
                 if (m_used == false) {
```
motowsk2: I would suggest simplifying this boolean check. Using !m_used is easier to read than comparing explicitly to false. This improves code clarity and follows common C++ style conventions. My suggestion would be to change this to:

if (!m_used)

source/tools/Random.hpp
```diff
@@ -107,135 +114,95 @@ namespace cse498 {
...
-                // Error handling: d_min must be less than d_max
-                if (d_min > d_max) {
-                    throw std::runtime_error("cse498::Random::GetDouble(): d_min must be less than d_max.");
-                }
 
-                check_rng();
+            /// @brief Templated function to generate and return values
+            /// @param The range of values to generate a number between. min must be <= max.
+            /// @return A random value in range of the specified type
+            template <typename T>
```
motowsk2: I would suggest constraining this template function to separate integral and floating-point behavior. Right now, it attempts to handle all types (including bool and char), which can lead to incorrect behavior and reduced type safety. Constraining the template improves correctness and makes the intent clearer. My suggestion would be to change this to:

template <std::integral T>
T GetValue(T min, T max);

template <std::floating_point T>
T GetValue(T min, T max);

source/tools/Random.hpp
```diff
@@ -107,135 +114,95 @@ namespace cse498 {
...
-            bool GetBool(){
-                //Generate between 0 and 1 using an int
-                int i = GetInt(0,1);
-
-                if (i == 0){
-                    return false;
+                // Turn the bits into the appropriate type
+                if constexpr (std::is_integral_v<T>) {
+                    // Handle int type values: ints, bools, chars
+                    return min + static_cast<T> (r % static_cast<uint64_t>(max-min+1));
```
motowsk2: I would suggest avoiding the use of modulo for generating bounded integers. Using % can introduce statistical bias when the range does not evenly divide the underlying random number space. This can lead to non-uniform distributions, which is especially important in simulations or games. A better approach would be to use rejection sampling or a scaling method to ensure uniformity.

source/tools/Random.hpp
```diff
@@ -107,135 +114,95 @@ namespace cse498 {
...
+                    return static_cast<T>(min + decimal_value * (max-min));
                 }
-                return true;
             }
 
-            // Generates based off of a given probability, and returns a bool
+            /// @brief Generates based off of a given probability, and returns a bool
+            /// @param the desired probaility of a true value
+            /// @return a weighted generated bool
             bool P(double probability = 0.5){
```
motowsk2: I would suggest adding [[nodiscard] where appropriate. This helps prevent accidental misuse of return values and communicates function intent more clearly. It also allows the compiler to perform better optimizations. My suggestion would be to change this to:

[[nodiscard]] bool P(double probability = 0.5);

source/tools/Random.hpp
motowsk2, line 1: ### **Implementation Quality Summary:**

Overall, this class demonstrates a solid foundation in implementation quality, with good structure and clear organization. The code is properly contained within the cse498 namespace and avoids namespace pollution, which is a strong positive. The logic is modular, with helper functions used to separate responsibilities such as state initialization, number generation, and type conversion, which improves readability and maintainability.

However, there are a few areas where implementation quality can be improved. The code may not compile cleanly under stricter flags like -Wall -Wextra due to missing headers (e.g., <cassert>, <type_traits>), and enabling -Werror would help catch these issues early. While the implementation generally follows the C++ standard, there are some areas that could be refined to better avoid potential issues, such as improving type safety in the templated GetValue function and avoiding biased integer generation.

There are opportunities to further adopt modern practices such as using std::array instead of raw arrays, range-based loops to reduce duplication, and more expressive type constraints. The use of standard library algorithms is somewhat limited, and some manual loops could be replaced with more idiomatic STL approaches.

There are also some naming inconsistencies, particularly with the use of the m_ prefix on types and helper structs (e.g., state structs), where m_ is typically reserved for member variables. This can make the code harder to read and distinguish between types and instance data, so aligning naming conventions would improve clarity.

Pointer handling is done safely overall, as there is no dynamic memory allocation or misuse of ownership, though avoiding unnecessary raw pointer aliases would further improve clarity. Const correctness is partially implemented (GetSeed), but could be expanded by marking additional functions as const or [[nodiscard]] where appropriate.

Overall, this is a well-structured implementation that meets many of the core requirements, but would benefit from improved adherence to modern C++ practices, stricter compilation standards, stronger type safety, and more consistent naming conventions to reach a higher level of quality.

### **Advanced C++ Features Summary:**

This class incorporates several advanced C++ features, particularly through the use of constexpr, templates, and value semantics. The use of constexpr for constants is well done and helps eliminate magic numbers while improving readability and potential compile-time optimization. Additionally, the templated GetValue function shows an effort to generalize behavior across multiple types, which is a strong step toward more flexible and reusable code. The class also makes good use of value semantics by storing RNG state directly within the object rather than relying on dynamic memory, which improves safety and simplicity.

However, there are some areas where the use of advanced C++ features could be strengthened. While templates are used, they are currently too broad and would benefit from stronger type constraints (separating integral and floating-point types). This would improve type safety and better demonstrate a deeper understanding of template design. Additionally, although modern constructs are introduced, they are not fully leveraged—for example, there is no use of lambdas or ranges, and opportunities exist to use range-based loops more consistently.

Overall, the class successfully incorporates at least two advanced C++ concepts (constexpr, templates, and value semantics), but could be improved by applying these techniques more rigorously and expanding into additional modern features. Strengthening template constraints and adopting more expressive modern patterns would elevate the implementation further.

source/tools/WeightedSet.hpp
```diff
@@ -68,24 +82,27 @@ namespace cse498 {
...
+			//Treat numbers below tolerance as 0
 			FixTinyNum(weight);
 
-			size_t idx = set_size;
-			item_idx[id] = idx;
-			items.push_back(id);
-			weights.push_back(weight);
-			sum_tree.push_back(0.0);
-			++set_size;
+			size_t idx = m_set_size;
```
motowsk2: I would suggest consistently using either size_t or std::size_t throughout the class rather than mixing both. While both refer to the same type, using one style consistently improves readability and makes the code feel more polished and uniform. My suggestion would be to standardize on:
std::size_t

source/tools/WeightedSet.hpp
```diff
@@ -68,24 +82,27 @@ namespace cse498 {
...
+			m_item_idx[id] = idx;
+			m_items.push_back(id);
+			m_weights.push_back(weight);
+			m_sum_tree.push_back(0.0);
+			++m_set_size;
```
motowsk2, line 137: I would suggest avoiding repeated lookups into m_item_idx after already verifying that the key exists. Repeated map access is slightly less efficient and also makes the code a little harder to follow. Using .at() or storing the iterator/value once makes the intent clearer and avoids any chance of accidental insertion if the logic changes later. My suggestion would be to change this to:

std::size_t idx = m_item_idx.at(id);

source/tools/WeightedSet.hpp
```diff
@@ -68,24 +82,27 @@ namespace cse498 {
 		 *
 		 * @throws std::invalid_argument if weight is negative or id already exists.
 		 */
-		void Insert(const T& id, double weight){
+		std::expected<void, std::string> Insert(const T& id, double weight){
```
motowsk2: I would suggest adding [[nodiscard]] to functions like Insert, Update, and Sample. Since these functions return std::expected, ignoring the return value could cause errors to be silently missed. Adding [[nodiscard]] makes the interface safer by encouraging callers to handle success or failure explicitly. My suggestion would be to change this to:

[[nodiscard]] std::expected<void, std::string> Insert(const T& id, double weight);
[[nodiscard]] std::expected<void, std::string> Update(const T& id, double weight);
[[nodiscard]] std::expected<T, std::string> Sample(double num) const;

source/tools/WeightedSet.hpp
```diff
@@ -37,14 +51,14 @@ namespace cse498 {
 		 */
 		void FixSum(size_t idx, double w_change) {
```
motowsk2: I would suggest using std::size_t consistently in helper function parameters as well. Since the member variables already use size-related types, matching them in helper functions improves consistency and avoids mixing styles within the same class. My suggestion would be to change this to:

void FixSum(std::size_t idx, double w_change)

source/tools/WeightedSet.hpp
```diff
@@ -11,23 +11,37 @@
 #pragma once
 
 #include <cassert>
-#include <stdexcept>
+#include <expected>
 #include <vector>
 #include <unordered_map>
```
motowsk2, line 20: I would suggest explicitly including <string> since the class uses std::string in multiple std::expected return types. Depending on indirect includes can make the code less portable, because another header might stop including <string> in the future. My suggestion would be to add include for string

source/tools/WeightedSet.hpp
```diff
@@ -11,23 +11,37 @@
...
+		// Small tolerance used to handle floating-point precision errors in weight comparisons
 		static constexpr double WEIGHTED_SET_TOL = 1e-12;
 
 		//Storage should be relatively small assuming T will be a string or int (probable for ids)
-		std::unordered_map<T, size_t> item_idx{};
-		std::vector<T> items{};
-		std::vector<double> weights{};
-		std::vector<double> sum_tree{}; 
-		std::size_t set_size = 0;
+		std::unordered_map<T, size_t> m_item_idx{};
```
motowsk2: I would suggest constraining or documenting the template requirements for T. Since this class stores T in an std::unordered_map, the type must be hashable and equality comparable, but that requirement is currently only implicit. Making those expectations clearer would improve usability and make the template design more robust.

source/tools/WeightedSet.hpp
motowsk2: I would suggest renaming FixTinyNum to something more descriptive, such as NormalizeTinyWeight or ClampTinyWeightToZero. The current name works, but it is somewhat vague and does not immediately communicate that the function is specifically normalizing very small floating-point weights. A more descriptive name would improve readability and make the purpose of the helper clearer.

source/tools/WeightedSet.hpp
```diff
@@ -166,25 +185,26 @@ namespace cse498 {
...
+			return m_items[idx];
 		}
 
-		size_t GetSize() const { return set_size; }
-		double GetItemSum(const T& item) const { return sum_tree.at(item_idx.at(item)); }
-		double GetWeight(const T& item) const { return weights.at(item_idx.at(item)); }
+		size_t GetSize() const { return m_set_size; }
+		double GetItemSum(const T& item) const { return m_sum_tree.at(m_item_idx.at(item)); }
+		double GetWeight(const T& item) const { return m_weights.at(m_item_idx.at(item)); }
+		double GetTotalWeight() const { return m_sum_tree.at(0); }
```
motowsk2: I would suggest adding [[nodiscard]] and noexcept to simple getter functions where appropriate. These functions are intended to return information without modifying state, and marking them this way communicates intent more clearly while also making accidental ignored results less likely. My suggestion would be to change these to something like:

[[nodiscard]] std::size_t GetSize() const noexcept { return m_set_size; }
[[nodiscard]] double GetTotalWeight() const { return m_sum_tree.at(0); }

source/tools/WeightedSet.hpp
```diff
@@ -125,20 +144,20 @@ namespace cse498 {
...
 		 *
-		 * @throws std::runtime_error if the set is empty.
-		 * @throws std::invalid_argument if num is outside the valid range.
+		 * @note The set can not be empty.
+		 * @note num is inside  the valid range [0, total_weight].
 		 */
-		T Sample(double num) const {
-			if (sum_tree.empty() || sum_tree[0] <= WEIGHTED_SET_TOL){
-				throw std::runtime_error("cse498::WeightedSet::Sample(): Cannot sample from an empty WeightedSet");
+		std::expected<T, std::string> Sample(double num) const {
```
motowsk2: I would note that the use of std::expected here is a strong design choice. It makes error handling explicit and fits these operations well, since insertion, update, and sampling can all fail in normal use. This is a good example of applying a modern C++ feature in a way that directly improves overall implementation quality.

source/tools/WeightedSet.hpp
motowsk2, line 1: ### **Implementation Quality Summary**

Overall, this class demonstrates a strong foundation in implementation quality, with clear organization, good encapsulation of state, and a modular design. The code is properly contained within the cse498 namespace and avoids namespace pollution, which is a strong positive. The use of helper functions such as FixSum and FixTinyNum helps separate core responsibilities and improves readability by breaking repeated logic into smaller, reusable pieces. The class also handles ownership safely by relying entirely on standard library containers rather than manual memory management.

There are also several areas where the implementation could be strengthened. While the file appears structurally clean, there are some opportunities to improve const correctness, type consistency, and overall robustness. For example, some uses of size_t and std::size_t are mixed, which can reduce consistency. There are also a few places where lookups are repeated unnecessarily, such as using m_item_idx[id] after already confirming the key exists, where .at() or an iterator-based lookup would communicate intent more clearly and avoid accidental insertion in future edits. Additionally, although the code uses helper functions well, some small style and defensive programming improvements would make the implementation more portable and maintainable under stricter compiler settings such as -Wall -Wextra -Werror.

### **Advanced C++ Features Summary**

This class demonstrates a solid use of several advanced C++ features, especially through its use of templates, constexpr, std::expected, and value semantics. The class template allows the container to store arbitrary item types, which is a meaningful and appropriate use of templates for reusable generic code. The use of constexpr for the tolerance constant is also a good choice, since it avoids magic numbers and makes the code more readable and maintainable. Additionally, the class uses value semantics effectively by storing its state directly in standard library containers rather than relying on dynamic allocation or pointer-based ownership.

One especially strong modern feature here is the use of std::expected for error handling. This is a good fit for operations like Insert, Update, and Sample, where failure is a normal possibility and should be represented explicitly rather than through exceptions. That choice improves clarity and makes the interface easier to use safely.

That said, while the class clearly incorporates multiple advanced C++ concepts, there are still some opportunities to go further. There is no meaningful use of lambdas or ranges in this implementation, and while those are not strictly necessary here, some standard algorithms or more expressive iteration patterns could help modernize the code further. In addition, the template is currently unconstrained, so the implementation assumes that T is hashable and equality-comparable without expressing those requirements directly. Adding constraints or documenting those assumptions more explicitly would strengthen the use of templates and show a deeper command of advanced C++ design.

Overall, this class does a good job incorporating at least three advanced C++ concepts in a meaningful way, particularly templates, constexpr, std::expected, and value semantics. The advanced features are not just included for appearance, but actually support the design of the class in a useful way.

source/tools/Random.hpp
GumdropHorse: This was just a personal preference choice, it has now been updated to better align with the company standards. Thanks!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Again, this was just a personal preference choice. It has now been updated to better align with the company standards. Thanks!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the suggestion, this has now been implemented. Thank you!

source/tools/Random.hpp
GumdropHorse: Thank you for the detailed feedback! Random has now been updated to fix everything you mentioned!

source/tools/Random.hpp
```diff
@@ -49,53 +51,50 @@ namespace cse498 {
...
+                struct Splitmix64State sm = {m_seed};
 
-                state.s[0] = splitmix64(sm);
-                state.s[1] = splitmix64(sm);
-                state.s[2] = splitmix64(sm);
-                state.s[3] = splitmix64(sm);
-            }
-
-            // Performs a left rotation on x by k bits
-            uint64_t rol64(uint64_t x, int k) {
```
JordanSebagh, line 79: Could be a more informative identifier than 'rol64'

source/tools/Random.hpp
```diff
@@ -49,53 +51,50 @@ namespace cse498 {
...
             // changes each time a number is generated
-            struct xoshiro256pp_state {
-                uint64_t s[STATE_NUMBER];
+            struct Xoshiro256ppState {
+                std::array<uint64_t, STATE_NUMBER> s;
             };
 
-            struct splitmix64_state {
+            struct Splitmix64State {
                 uint64_t s;
```
JordanSebagh: Could be a more informative identifier than 's'

source/tools/Random.hpp
JordanSebagh, line 79: use more informative names

source/tools/Random.hpp
```diff
@@ -49,53 +51,50 @@ namespace cse498 {
...
+            /// @brief Performs a left rotation on x by k bits
+            /// @param x and k, x is the value being rotated and k is how much it is rotated by
+            /// @return A rotated x value
+            uint64_t Rol64(uint64_t x, int k) {
                 return (x<<k) | (x >> (NUM_BITS-k));
             }
 
-            // Generates a random number
-            uint64_t xoshiro256pp(struct xoshiro256pp_state &state) {
-                uint64_t *s = state.s;
```
JordanSebagh, line 94: Could be a more informative identifier than 's'

source/tools/Random.hpp
```diff
@@ -107,135 +106,100 @@ namespace cse498 {
...
                 return result;
             }
 
             ///
             ////////////////////////////////////////////////////////////
 
-            
-            // Generates a double
-            double double_xoshiro(struct xoshiro256pp_state &state) {
-                uint64_t r = xoshiro256pp(state);
```
JordanSebagh, line 120: Could be a more informative identifier than 'r'

source/tools/Random.hpp
JordanSebagh, line 1: You did a good job moving many numeric values into named constants instead of scattering numbers throughout the code. That made the implementation easier to read

JordanSebagh, line 1: The constructor of this class does not actually take an optional seed, even though the original class description says it should as seen on the GitHub description of group 15 classes. Right now there is only a default constructor, and the seed must be changed later with SetSeed

JordanSebagh, line 1: Good job on having lots of informative comments. One thing to note is that you have some spelling errors in some areas in the comments such as “Ensurses,” “initalized,” “Allosw,” and “probaility.”

source/tools/Random.hpp
```diff
@@ -49,53 +51,50 @@ namespace cse498 {
...
+            /// @return A well-mixed, non-zero uint64_t value
+            uint64_t Splitmix64(struct Splitmix64State &state) {
                 uint64_t result = (state.s += GOLDEN_RATIO);
                 result = (result ^ (result >> RIGHT_SHIFT1)) * FIRST_MIXING;
                 result = (result ^ (result >> RIGHT_SHIFT2)) * SECOND_MIXING;
                 return result ^ (result >> RIGHT_SHIFT3);
             }
 
-            // Uses m_seed to generate the state positions
-            void xoshiro256pp_init(struct xoshiro256pp_state &state) {
```
JordanSebagh, line 69: Style is mostly good, however, there is a mix between Doxygen comments and more casual inline comments.

source/tools/Random.hpp
JordanSebagh, line 1: The visual appearance is solid. Overall, it is readable. The file is split into logical sections and the public/private structure is clear.

source/tools/WeightedSet.hpp
```diff
@@ -11,23 +11,37 @@
...
+
+		// Small tolerance used to handle floating-point precision errors in weight comparisons
 		static constexpr double WEIGHTED_SET_TOL = 1e-12;
 
 		//Storage should be relatively small assuming T will be a string or int (probable for ids)
-		std::unordered_map<T, size_t> item_idx{};
-		std::vector<T> items{};
-		std::vector<double> weights{};
-		std::vector<double> sum_tree{}; 
-		std::size_t set_size = 0;
```
JordanSebagh, line 30: There is a small mix of size_t and std::size_t. It would be better to pick one style and use it consistently.

source/tools/WeightedSet.hpp
JordanSebagh, line 1: The WeightedSet class has a strong overall design. I thought the class level comments were especially helpful because they explain both what the class does and how the vector-based binary sum tree supports efficient updates and sampling. Naming is mostly good, except a few identifiers like FixTinyNum, FixSum, and the parameter name num could be more descriptive. The code also avoids magic numbers. The biggest interface issue is that Sample(double num) does not actually generate a random sample itself, so the function name may be a little misleading unless the caller already knows they must provide a cumulative sample value. Overall, the design is solid and readable, with only a few naming and interface clarity improvements needed. You also have a few spelling errors in the comments such as 'inverval'. The class makes good use of advanced C++ features such as templates and std::expected

source/core/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+struct LinkedRooms {
+    int x1, y1; //first room's coordinates
+    int x2, y2; //second room's coordinates
+};
+
+namespace cse498 {
+    
+
+    class WorldGen {
```
JordanSebagh: WorldGeneration.hpp defines a class named WorldGen while the file name says “WorldGeneration”

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+#pragma once 
+
+
+#include "BSP-Dungeon.hpp"
+#include <cmath>
+#include <ranges>
+#include <algorithm>
+
+struct Point {
```
JordanSebagh, line 21: Point and LinkedRooms sit outside namespace cse498

source/core/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,384 @@
...
+        void RepopulateTree() { 
+            insert_split(m_iterations); //Creates BSP Tree
+            PostOrderDFS(); //Grabs all the generated room slots from the tree
+        }
+
+        /**
+         * @brief Regenerates the BSP Tree incase any modifications to width/height/properties after its creation are made in order to 
+         * ensure the object's state is up to date when new dungeons are later created
+         */
+        void RegnerateObjectState() {
```
JordanSebagh: RegnerateObjectState in is a typo (Regenerate).

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,384 @@
...
+        BSPNode() = default;
+        BSPNode(int l, int r, int x, int y, int width, int height, std::string name) 
+           : left_child(l), right_child(r), x(x), y(y), width(width), height(height), name(name) {}
+
+    };
+    
+
+    static constexpr int DEFAULT_WIDTH = 150;
+    static constexpr int DEFAULT_HEIGHT = 100;
+    static constexpr int DEFAULT_WIDTH_THRESHOLD = 30;
```
JordanSebagh, line 50: Good job here to avoid magic numbers

source/Worlds/DungeonOne.hpp
```diff
@@ -0,0 +1,35 @@
...
+    private:
+    static cse498::WeightedSet<std::string> MakeRoomPool() {
+      cse498::WeightedSet<std::string> rooms;
+		  std::string file_name = "one_pool/room_";
+      rooms.Insert(file_name + "1.txt", 10);
+      rooms.Insert(file_name + "2.txt", 10);
+      rooms.Insert(file_name + "3.txt", 10);
+      rooms.Insert(file_name + "4.txt", 10);
+      rooms.Insert(file_name + "5.txt", 10);
+      rooms.Insert(file_name + "6.txt", 1);
```
JordanSebagh: Hardcodes room weights (10 vs 1 for room 6) with no named constants

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+            int level = 1; // TODO: Add a counter for number of levels generated
+            std::string file_path = m_image_path;
+
+            // World objects (floor tiles, walls, doors, barriers, ect.)
+            if (tile_c == '&' || tile_c == '^' || tile_c == '<' || tile_c == '>' || tile_c == '$'
+                || tile_c == 'd' || tile_c == 's' || tile_c == ' ' || tile_c == 'v') {
+
+                // Level folder
+                file_path += "world/";
+                if (level == 1) {file_path += "forest/";}
```
JordanSebagh: Uses level == 1,2,3 and m_rng.GetValue(2,5) without names. Would prefer something like MIN_FLOOR_VARIANT, MAX_FLOOR_VARIANT, or an enum.

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,384 @@
...
+            return m_leaf_nodes;
+        }
+
+        /// @brief Returns the entirety of the created BSP_Tree
+        [[nodiscard]] std::vector<BSPNode> GetBSPTree() const {
+            return m_BSP_tree;
+        }
+
+        /// @brief Returns the reference to BSP_Tree's leaf nodes
+        [[nodiscard]] std::vector<BSPNode>& GetLeafNodes() {
```
JordanSebagh, line 248: GetLeafNodes() / GetBSPTree() exist as both const and non const. Try to be consistent

source/Worlds/Dungeon/DungeonBase.hpp
```diff
@@ -0,0 +1,103 @@
...
+      //generation.GetBSP().GenerateTileMap();
+
+      main_grid.Load(testing);
+
+
+    }
+    ~DungeonBase() = default;
+
+    /// Allow the agents to move around the maze.
+    int DoAction(AgentBase & agent, size_t action_id) override {
```
JordanSebagh, line 73: This function is declared int but uses return false / return true in places. This works but slightly misleading.

source/core/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
+/**
+ * This file is part of the Fall 2026, CSE 498, section 2, course project.
+ * Attributed References for the 'random_splits' function: https://eskerda.com/bsp-dungeon-generation/
+ * References: https://www.roguebasin.com/index.php/Basic_BSP_Dungeon_generation 
+ * References: https://en.wikipedia.org/wiki/Binary_space_partitioning
+ * @brief A BSP tree
```
JordanSebagh: “@brief A BSP tree” — copied from the wrong file it seems like

source/Worlds/Dungeon/RoomHolder.hpp
JordanSebagh, line 1: Several Doxygen blocks are empty

source/core/DungeonBase.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include "../tools/WeightedSet.hpp"
+
+namespace cse498 {
+
+  class DungeonBase : public WorldBase {
+  protected:
+    enum ActionType { REMAIN_STILL=0, MOVE_UP, MOVE_DOWN, MOVE_LEFT, MOVE_RIGHT };
+
+    size_t floor_id; ///< Easy access to floor CellType ID.
+    size_t wall_id;  ///< Easy access to wall CellType ID.
```
JordanSebagh: Does not adhere to the 'm' member variable prefix for all these member variables

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+ #pragma once
+
+ #include <vector>
+ #include <string>
+ #include <cmath>
+ #include <fstream>
+ #include <iostream>
+ #include <cassert>
+ #include <iostream>
```
JordanSebagh: Duplicate includes of iostream

demos/GenerativeWorld/Group15_main.cpp
JordanSebagh, line 1: Group15_demo.cpp is a solid driver as it includes agents and interfaces and runs a DungeonOne with agents. This does a solid job of showing how the generative world plugs into shared classes that we already have as well as how your world functions and looks.

group_specific_content/Group-15/Group 15 Module - Generative World.md
JordanSebagh, line 1: Good job providing documentation on your classes and specifically how to build and run your program. I also liked how you explain the character meanings, as that can be confusing if we were just looking at the world. I also like how you fully explained all classes and their functionality.

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    public:
+		RoomHolder(const cse498::WeightedSet<std::string>& room_pool,
+				   const std::string& file_path) 
+			: m_file_path(file_path),
+			  m_rng(),
+			  m_room_pool(room_pool)
+		{}
+
+        /// @brief
```
Lak453: Several @brief and @return comments are left blank. Filling those in would make the interface much easier for another developer to understand.

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+ **/
+
+
+ #pragma once
+
+ #include <vector>
+ #include <string>
+ #include <cmath>
+ #include <fstream>
+ #include <iostream>
```
Lak453: Duplicate includes

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+		RoomHolder(const cse498::WeightedSet<std::string>& room_pool,
+				   const std::string& file_path) 
+			: m_file_path(file_path),
+			  m_rng(),
+			  m_room_pool(room_pool)
+		{}
+
+        /// @brief 
+        /// @return 
+        [[nodiscard]] std::vector<std::string> GetCurrentRoom() {
```
Lak453: If the intent is only to read the currently loaded room, returning const std::vector<std::string>& would avoid an unnecessary copy and make the interface more efficient.

source/tools/Random.hpp
```diff
@@ -107,135 +106,100 @@ namespace cse498 {
...
 
-            // Generates based off of a given probability, and returns a bool
-            bool P(double probability = 0.5){
+            /// @brief Generates based off of a given probability, and returns a bool
+            /// @param the desired probaility of a true value
+            /// @return a weighted generated bool
+            [[nodiscard]] bool P(double probability = 0.5){
                 // Error handling: probability must be between 0 and 1
                 if (0 > probability || 1 < probability) {
                     throw std::runtime_error("cse498::Random::P(): parameter probability must be between 0 and 1.");
```
tmaks2: As a company we decided to avoid exceptions unless absolutely necessary. Additionally, the class kind of mixes error-handling styles since both GetValue() overloads use assert while P() throws for invalid inputs.

source/core/RoomHolder.hpp
Lak453: [[nodiscard]] is used in several getter-style functions such as GetCurrent Room(), GetRoomWidth(), and Generate File Path(). That is a good use of modern C++.

source/core/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+                }
+
+            }
+        }
+
+        
+        /// @brief Calculate center of room placed in grid
+        /// @param room we're inputting the BSP_Tree Node's room value
+        /// @return pair of coordinates 
+        [[nodiscard]] Point CalcRoomCenter(std::vector<std::string>& room) {
```
Lak453: Does CalcRoomCenter(std::vector<std::string>& room) not modify room? If so, the parameter should probably be const std::vector<std::string>& room.

source/core/WorldGeneration.hpp
Lak453: The naming is kind of messy

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+    int x, y; // x-y points of a room
+};
+
+struct LinkedRooms {
```
Lak453, line 25: Why  LinkedRooms, Point is outside of cse498?  What is the reason?

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+                point_x = std::abs(point_x);
+            }
+
+            for (int y = 0; y <= point_y; ++y) {
+                int y_point;
+                if (negative_y) y_point = y1_value - y;
+                else y_point = y1_value + y;
+
+                auto &y_char = m_grid[y_point][x1_value];
+                auto it = std::ranges::find(wall_set, y_char);
```
Lak453, line 164: Using std::ranges::find instead of a manual loop is a nice use of newer C++ library features.

tests/tools/GenerativeWorld/Random.cpp
tmaks2, line 1: Overall, the Random tests seem comprehensive. The reproducibility checks are strong and do a great job validating deterministic behavior through seeding

source/tools/WeightedSet.hpp
```diff
@@ -68,24 +82,27 @@ namespace cse498 {
 		 *
 		 * @throws std::invalid_argument if weight is negative or id already exists.
```
tmaks2: The implementation now uses std::expected which is good with the non-exception policy we decided on as a company, but your doc comments for Insert() and Update() still say that it throws. I would update the comments so the documentation matches the current implementation.

tests/tools/GenerativeWorld/WeightedSet.cpp
tmaks2, line 1: Overall, this test file for WeightedSet is pretty strong. It does a good job covering both the normal behavior and error cases for Insert(), Update(), and Sample().

source/tools/WeightedSet.hpp
```diff
@@ -166,25 +185,26 @@ namespace cse498 {
...
 			}
-			return items[idx];
+			return m_items[idx];
 		}
 
-		size_t GetSize() const { return set_size; }
-		double GetItemSum(const T& item) const { return sum_tree.at(item_idx.at(item)); }
-		double GetWeight(const T& item) const { return weights.at(item_idx.at(item)); }
+		size_t GetSize() const { return m_set_size; }
+		double GetItemSum(const T& item) const { return m_sum_tree.at(m_item_idx.at(item)); }
```
tmaks2: GetItemSum(), GetWeight(), and GetTotalWeight use .at(), which I believe can throw on invalid access. You could either possibly document these accessors as being more strict pre-condition functions and use asserts or convert them to return std::expected like Insert(), Update(), and Sample() do.

source/core/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        /// @brief Loads a random pre-made room based off of the level dungeon we're currently in
+        /// @return 
+        std::vector<std::string> LoadRoom() {
+            std::string selected_pool = GenerateFilePath();
+            std::ifstream file(m_file_path + selected_pool); // this will open one of the rooms
+            //std::cout << m_file_path + selected_pool << std::endl;
+
+            assert(file.is_open());
```
tmaks2: I think this would be better handled as a recoverable error path rather than an assertion, since missing or misconfigured files are runtime failures that can occur during normal execution. In release builds, the assertion may be removed entirely, which would leave this failure unchecked.

source/Worlds/Dungeon/RoomHolder.hpp
```diff
@@ -0,0 +1,217 @@
...
+            auto height_midpoint = std::ceil(GetRoomHeight() / 2);
+
+            return std::make_pair(width_midpoint, height_midpoint);
+        }
+
+        [[nodiscard]] std::string GenerateFilePath() { 
+			auto room_select = m_rng.GetValue(0.0, m_room_pool.GetTotalWeight());
+			
+			auto sample_result = m_room_pool.Sample(room_select);
+			assert(sample_result.has_value());
```
tmaks2, line 87: Sample() returns a std::expected, but the result is asserted instead of handled. This bypasses the intended error-handling design and may turn a recoverable failure into a crash, especially since assertions can be removed in release builds.

source/Worlds/DungeonOne.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "../core/DungeonBase.hpp"
+
+namespace cse498 {
+
+  class DungeonOne: public DungeonBase {
+    private:
+    static cse498::WeightedSet<std::string> MakeRoomPool() {
+      cse498::WeightedSet<std::string> rooms;
+		  std::string file_name = "one_pool/room_";
+      rooms.Insert(file_name + "1.txt", 10);
```
tmaks2: These Insert() calls ignore the return std::expected which means that room-pool setup can fail silently even though the WeightedSet API is designed to report those failures.

source/core/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+        /// @brief grabs width of the rid map
+        /// @return WIDTH value
+        [[nodiscard]] size_t GetWidth() {
+            return m_width;
+        }   
+
+        /// @brief updates the HEIGHT value for the grid map
+        /// @param height 
+        void SetHeight(size_t height) {
```
tmaks2: SetHeight(), SetWidth(), and SetIterations() all take size_t, but the underlying member variables are stored as int. This creates a type mismatch and it may be cleaner to keep the parameter and member types consistent

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,187 @@
...
+        CHECK(tree.size() == 29);
+        CHECK(leaf.size() == 15);
+
+
+        //Testing with Seed value 23456
+        //Refer to BSP-Dungeon-Seed-Info/BSP-Dungeon-Seed-23456.md for Tree information
+        BSP.SetRngSeed(SEED_VALUE_TWO);
+        CHECK(BSP.GetRngSeed() == SEED_VALUE_TWO);
+        BSP.RegnerateObjectState();
+        CHECK(tree.size() == 29);
```
tmaks2: I think GetLeafNodes() and GetBSPTree() create copies rather than give a reference, so leaf and tree are copied before regeneration, so these later CHECKs are still validating the original snapshot rather than the regenerated BSP state. These should probably be re-read from BSP after each RegnerateObjectState() call so the test actually verifies the updated tree and leaf data.

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../../../source/core/BSP-Dungeon.hpp"
+
+/// @brief Brought over from DungeonOne for test case purposes
+/// @return vector of 
+static cse498::WeightedSet<std::string> MakeRoomPool() {
+    cse498::WeightedSet<std::string> rooms;
+    std::string file_name = "one_pool/room_";
+    rooms.Insert(file_name + "1.txt", 10);
```
tmaks2: The room pool setup ignores all the returned std::expected values from Insert(). Since this is a test setup, I think it would be better to check that each insert succeeded so the test doesn't silently continue with an invalid setup.

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
tmaks2, line 1: These tests do a decent job for checking tree/leaf count stability across seeds, but the structure is currently verified mostly through container sizes. I think the tests would be stronger if they also validated additional properties of the generated layout, such as room placement, split correctness, or other structural characteristics beyond just node counts.

source/core/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+            m_connected_rooms.push_back(LinkedRooms{left.x, left.y, right.x, right.y}); //x1,y1,x2,y2 respectively
+
+            auto gamble = m_rng.GetValue(0,1); //Determines which node is returned
+
+            //sends a node upwards, allowing connectivity between nodes for linking 
+            if (gamble == 0) {
+                return right; 
+            }
+            else {
```
tmaks2: This else statement is redundant. Either remove the `return left;` at the end or this else block.

source/core/WorldGeneration.hpp
```diff
@@ -0,0 +1,214 @@
...
+        std::vector<std::string> m_grid; //Grid we're rasterizing information from mBSP_Tree to
+        std::vector<LinkedRooms> m_connected_rooms; //The x-y coord pairs of two rooms used for connecting the room pair
+        Random m_rng;
+        int m_offset_x; //x offset of room placement
+        int m_offset_y; //y offset of room placement
+            
+    public: 
+
+        /// @brief Creates and initializes BSP Tree, RoomHolder, and grid for outputting dungeon level
+        WorldGen(const cse498::WeightedSet<std::string>& room_pool)
```
tmaks2: I don't need the cse498 qualifier is needed since the class is in the cse498 namespace.

source/Worlds/Dungeon/DungeonBase.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+      WorldGen generation(room_pool); 
+      generation.CreateDungeon(); 
+      std::vector<std::string> testing = generation.GetDungeon();
+
+      // Debugging calls in order to see the Grid Parition outline and coordinate/room information
+      //generation.GetBSP().TreeParser();
+      //generation.GetBSP().GenerateTileMap();
+
+      main_grid.Load(testing);
```
tmaks2, line 66: The dungeon is generated and then loaded, which is good, but I think there should be some validation of the generated layout before calling `main_grid.Load(testing);`. I think it would be helpful to either validate the generated dungeon or expose a clear failure path since world generation depends on external room data and some earlier setup steps.

tests/tools/GenerativeWorld/Random.cpp
Deni-Tepic, line 1: strong test cases but it could be useful to add tests for invalid inputs as well, such as min > max for GetValue() and out-of-range values for P(),

just to border on the side of caution

demos/GenerativeWorld/Group15_main.cpp
tmaks2, line 1: Overall, this module was fairly strong in both implementation and testing. The revised Random and WeightedSet classes make good use of modern C++ features like std::expected and concepts, and the Random tests did a good job with validating deterministic behavior through seeding.

For implementation quality, I think the biggest weaknesses are around error handling consistency and some assumptions about successful setup. There are a few places where recoverable runtime failures are handled with assertions or where std::expected results are returned but not checked. I also noticed some of the broader module files are getting fairly implementation heavy in header files, so it might be good to split some of that logic into corresponding .cpp files to improve readability, compile-time hygiene, and maintainability.

For testing and error checking, I think the test coverage is strongest for the revised classes and a little weaker for some of the broader dungeon generation behavior. When I ran the demo, I also noticed at least one case where a path to another room was blocked on one side by internal room obstacles. It might be worth adding some validation or generation constraints to prevent room connections from being partially blocked like that, since that seems like a kind of issue that could slip through if tests focus more on node counts than actual playability or traversability.

Deni-Tepic (commented):
+ I like how you split up the integral and floating-point versions of GetValue() -- it makes it a lot easier to follow what’s going on in each case.

+ The seed-based tests are nice.  It is really easy to check reproducibility.

+ TagManager test case: The coverage here is pretty thorough overall, impressive!  Especially the include/exclude query cases and the duplicate add behavior.

- Randomhpp Tests OVERALL look good-- could be nice to also cover some invalid input cases like bad probability values or min > max.

- Random.hpp implementation:
        - Might be helpful to clarify whether the floating-point version of GetValue()  
        returns values in [min, max] or [min, max).
         - There’s a mix of error handling styles here (assert in GetValue() vs exceptions in 
        P()), so it might be worth standardizing that behavior.

- LINE 326 of  TestPathGenerator - Assertion is checking trapped instead of trapped2.
    "// edge case
    auto trapped2 = PathGenerator::FindShortestPath({1, 2}, {1, 2}, request2_1);
    CHECK(!trapped);"
    
- TestWorldPath: Might also be worth testing Extend when one of the paths is empty, just to cover those edge cases.



*Overall this was a solid module to review. The demo made it easy to see what was going on, and it’s clear a lotttt of work went into both the implementation and the tests. 

I left a few smaller comments, but overall this looks like a strong contribution to the project. 

Feel free to reach out to me on here or Discord if you want to discuss any of my reviews.

motowsk2 (commented):
Submitting the comments I left for review, forgot to do this when I left them a couple of days ago. I was responsible for reviewing Advanced C++ features and Implementation Quality within the Random.hpp and WeightedSet.hpp files.

source/tools/Random.hpp
GumdropHorse: Updated! Thanks for the feedback!

source/tools/Random.hpp
GumdropHorse: Updated! Thanks for the feedback!

source/tools/Random.hpp
GumdropHorse: Updated! Thanks for the feedback!

source/tools/Random.hpp
GumdropHorse: Updated! Thanks for the feedback!

source/tools/Random.hpp
GumdropHorse: Updated! Thanks for the feedback!

source/tools/Random.hpp
GumdropHorse: This has been fixed, thanks for the feedback!

GumdropHorse: Updated! Thanks for catching this!

source/tools/Random.hpp
GumdropHorse: Thanks for catching this! Fixed!

tests/tools/GenerativeWorld/Random.cpp
GumdropHorse: Good catch, fixed!

source/Worlds/Dungeon/WorldGeneration.hpp
joeyhyun1: Fixed

source/core/BSP-Dungeon.hpp
joeyhyun1: Fixed

source/tools/WeightedSet.hpp
AbbyFranz: Fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: Thank you!

tests/tools/GenerativeWorld/WeightedSet.cpp
AbbyFranz: Thank you!

source/tools/WeightedSet.hpp
AbbyFranz: Fixed

source/tools/WeightedSet.hpp
AbbyFranz: I'm not sure what you mean about repeated lookups here. A lookup only happens once with .contains() and then if the item does not exist it is inserted using operator[]. Since the item doesn't exist within the map yet using .at() would cause an exception. 

However, I do appreciate the advice, since this prompted me to look for duplicate map lookups and I found one in a function further down.

source/Worlds/Dungeon/WorldGeneration.hpp
paulbui081: Fixed!

source/tools/WeightedSet.hpp
AbbyFranz: fixed

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/core/WorldGeneration.hpp
paulbui081: fixed

source/core/WorldGeneration.hpp
paulbui081: fixed

source/core/WorldGeneration.hpp
paulbui081: fixed!

source/core/WorldGeneration.hpp
paulbui081: Fixed

source/tools/Random.hpp
GumdropHorse: This has been updated, thank you for the feedback!

source/core/WorldGeneration.hpp
paulbui081: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: fixed

source/tools/WeightedSet.hpp
AbbyFranz: I changed the names for FixTinyNum, FixSum, and the parameter name num. I left sample as is because it's still technically sampling randomly, as it takes the random value for sampling as a parameter instead of generating it within, but it would still work the same either way.

AbbyFranz: Thank you for your thorough review!

source/Worlds/DungeonOne.hpp
AbbyFranz: fixed, room weights were moved into a single container with documentation explaining individual numbers

source/Worlds/Dungeon/RoomHolder.hpp
AbbyFranz: We're using assert here intentionally because std::unexpected occuring here would mean that the set is either empty, an invalid value was sent into Sample(), or there is a bug within Sample().  At this point, the room pool should already be valid and the sampled value should generated from [0, total_weight], so Sample() shouldn't fail unless there is a bug in pool construction, in the RNG range generation, or in Sample() itself. This also means the error is not recoverable. If the internal logic is already incorrect, retrying or using a fallback would only hide the bug, or possibly still result in a bug. Therefore we felt it appropriate to use assert here instead.

source/core/RoomHolder.hpp
paulbui081: fixed

source/core/RoomHolder.hpp
paulbui081: fixed

source/core/RoomHolder.hpp
paulbui081: fixed

source/core/RoomHolder.hpp
paulbui081: fixed

source/Worlds/Dungeon/DungeonBase.hpp
paulbui081: This function overrides the abstract virtual DoAction(...) found in instructor's version of WorldBase, which is set up to use int as a true/false flag. We could change it, but for compatibility reasons with everyone else's already existing code, we're opting to keep it as it is.

source/core/BSP-Dungeon.hpp
paulbui081: fixed

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: fixed

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
paulbui081: Thank you for pointing this out! This has been fixed

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
paulbui081: Thank you for the tips! Beyond what's already established, I'll look into adding these additional BSP tests

source/core/RoomHolder.hpp
GumdropHorse: Thank you! This has been fixed.

source/Worlds/Dungeon/DungeonBase.hpp
paulbui081: Thanks for the tip! We'll look into adding validation to ensure that it properly generates.

source/Worlds/Dungeon/DungeonBase.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include "WorldGeneration.hpp"
+#include "../tools/WeightedSet.hpp"
+
+namespace cse498 {
+
+  class DungeonBase : public WorldBase {
+  protected:
+    enum ActionType { REMAIN_STILL=0, MOVE_UP, MOVE_DOWN, MOVE_LEFT, MOVE_RIGHT };
+
+    size_t m_floor_id; ///< Easy access to floor CellType ID.
```
karensuzue, line 21: This does not follow the company standard for member variable names

source/tools/WeightedSet.hpp
```diff
@@ -11,44 +11,88 @@
...
+	/**
+	 * @concept UnorderedMapKey
+	 * @brief A type that can be used as a key in std::unordered_map.
+	 *
+	 * Requires that T is hashable and equality comparable.
+	 *
+	 * @tparam T The type being checked.
+	 */
 	template <typename T>
+	concept UnorderedMapKey = Hashable<T> && std::equality_comparable<T>;
```
karensuzue, line 47: Very good use of concepts!!

source/tools/WeightedSet.hpp
```diff
@@ -11,44 +11,88 @@
...
+		// Small tolerance used to handle floating-point precision errors in weight comparisons
 		static constexpr double WEIGHTED_SET_TOL = 1e-12;
 
 		//Storage should be relatively small assuming T will be a string or int (probable for ids)
-		std::unordered_map<T, size_t> item_idx{};
-		std::vector<T> items{};
-		std::vector<double> weights{};
-		std::vector<double> sum_tree{}; 
-		std::size_t set_size = 0;
+		std::unordered_map<T, std::size_t> m_item_idx{};
```
karensuzue, line 70: Once again, you should update the variable names to match the company standard (mPascalCase)

source/tools/WeightedSet.hpp
```diff
@@ -11,44 +11,88 @@
...
+		std::vector<T> m_items{};
+		std::vector<double> m_weights{};
+		std::vector<double> m_sum_tree{}; 
+		std::size_t m_set_size = 0;
```
karensuzue, line 74: I think some of these variable names aren't the most intuitive or explanatory. I'd at least include a comment explaining what they are

source/tools/WeightedSet.hpp
```diff
@@ -11,44 +11,88 @@
...
 		}
 
 		/**
-		 * Normalizes very small weights to zero.
+		 * Changes all very small weights to be zero.
 		 * 
 		 * @param weight Weight value to normalize (modified in place).
 		 */
-		void FixTinyNum(double& weight){
+		void NormalizeTinyWeight(double& weight){
```
karensuzue, line 95: This might be nitpicky, but I don't think "normalize" is the right word here. This is more like a "clamp" to zero. But honestly it doesn't matter much :)

source/tools/WeightedSet.hpp
```diff
@@ -66,26 +110,35 @@ namespace cse498 {
...
+		 * @return std::expected<void, std::string>
+ 		 *         - On success: contains no value.
+ 		 *         - On failure: contains an error message describing the issue.
+		 *
+		 * @note Returns an error if:
+		 *       - weight is negative
+		 *       - id already exists in the set
 		 */
-		void Insert(const T& id, double weight){
+		[[nodiscard]] std::expected<void, std::string> Insert(const T& id, double weight){
```
karensuzue, line 121: Good use of expected here

source/tools/WeightedSet.hpp
```diff
@@ -121,70 +186,111 @@ namespace cse498 {
...
-		 * @throws std::invalid_argument if num is outside the valid range.
+		 * @note The set can not be empty.
+		 * @note sample_val is inside  the valid range [0, total_weight].
 		 */
-		T Sample(double num) const {
-			if (sum_tree.empty() || sum_tree[0] <= WEIGHTED_SET_TOL){
-				throw std::runtime_error("cse498::WeightedSet::Sample(): Cannot sample from an empty WeightedSet");
+		[[nodiscard]] std::expected<T, std::string> Sample(double sample_val) const {
+			if (m_sum_tree.empty() || m_sum_tree[0] <= WEIGHTED_SET_TOL){
+				return std::unexpected("cse498::WeightedSet::Sample(): Cannot sample from an empty WeightedSet");
```
karensuzue, line 198: Well it's technically empty set OR total weight too small, but they're effectively the same :)

source/tools/WeightedSet.hpp
```diff
@@ -121,70 +186,111 @@ namespace cse498 {
...
 			//lower and upper endpoints of the full range of values. Ex; [0.0, 5.1]
 			double outer_lo = 0.0; //lower value in outer interval
-			double outer_up = sum_tree[0]; //upper value in outer interval
-			size_t idx = 0;
+			double outer_up = m_sum_tree[0]; //upper value in outer interval
+			std::size_t idx = 0;
 
 			while(true){
-				assert(num >= outer_lo - WEIGHTED_SET_TOL && num <= outer_up + WEIGHTED_SET_TOL);
+				assert(sample_val >= outer_lo - WEIGHTED_SET_TOL && sample_val <= outer_up + WEIGHTED_SET_TOL);
```
karensuzue, line 210: I would include a message here explaining what this assert is checking for

tests/tools/GenerativeWorld/WeightedSet.cpp
karensuzue, line 1: You have very strong test cases here, especially for sampling. A few additional scenarios I would test separately are sampling from a very skewed tree, and sets whose sizes do not form a full binary tree (though it looks like you've already covered that in the final two test cases) :)

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        ////////////////////////////////////
+        //    BSP Tree management
+        ///////////////////////////////////
+
+        /// @brief Creates a split in the grid to populate with a room
+        /// @param iter number of iterations desired in the grid 
+        /// @return integer value of a BSPNode's position in its container
+        int insert_split(int iter) {
```
karensuzue, line 101: Ah, so this is the public starting point to populate it with a root node. Then it calls an overloaded version of the function below which continues to build out the tree. Nice!

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+    private:
+
+        /// @brief Continuously populates the BSP tree until iter base case reaches 0
+        /// @param iter number of iterations to split
+        [[nodiscard]] int insert_split(BSPNode node, int iter) {
+            int node_index = m_BSP_tree.size();
+            m_BSP_tree.push_back(node);
+
+            if (iter != 0) {
```
karensuzue, line 266: I wonder if there's a reason you're using int here instead of size_t for iter. If using int is necessary, I'd make this condition `if (iter > 0)`

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+                //connecting the nodes in the array-based tree based on index position
+                m_BSP_tree[node_index].left_child = left_node;
+                m_BSP_tree[node_index].right_child = right_node;
+
+            }
+            return node_index;
+        }
+
+        /// @brief Given an inputted struct root Node, splits the width/height randomly into  
+        /// @param node
```
karensuzue, line 293: The function takes an additional `iter`, so you should mention it here in the list of params :)

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+        /// @brief Given an inputted struct root Node, splits the width/height randomly into  
+        /// @param node 
+        /// @return returns a tuple pair of Nodes, other returning nullopt if threshold width/height not met
+        [[nodiscard]] std::optional<std::tuple<BSPNode, BSPNode>> random_split(BSPNode &node, int iter) {
+
+            bool split_width = true;
+            bool split_height = true;
+
+            //If the width or the height of the partition do not meet the minimum threshold, stop the split
```
karensuzue, line 300: I think the documentation could be clearer here, maybe something like: "Make sure the partition can be divided into two parts while still meeting the minimum threshold"

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+            BSPNode left_split, right_split;
+            bool directional_split;
+
+            if (split_width && split_height) {
+                directional_split = (m_rng.GetValue(0,1).value() == 0);
+            }
+            else {
+                directional_split = split_width;
+            }
```
karensuzue, line 321: Okay, this block worried me for a second but I see what you're doing. Maybe it would be clearer and safer to write it like this instead, even if it's a bit more work:
```
bool directional_split;
if (split_width && split_height) {
    directional_split = (m_rng.GetValue(0, 1).value() == 0);
} else if (split_width) {
    directional_split = true;
} else {
    directional_split = false;
}
```

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -0,0 +1,374 @@
...
+        /// @return integer value of a BSPNode's position in its container
+        int insert_split(int iter) {
+            BSPNode root_node = BSPNode{-1, -1, 0 , 0, m_width, m_height, std::string("Container")}; // creation of root node
+
+            return insert_split(root_node, iter);
+        }
+
+        /// @brief Grabs all the leaf nodes of the BSP_Tree using Post-Order DFS, for map generation
+        /// @param node root_node 
+        void PostOrderDFS(BSPNode &node) {
```
karensuzue, line 109: Yeah I think if the documentation was a little more specific, that would be great ! From what I understand, this grabs rooms and applies them to each of the leaf nodes in the tree

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,257 @@
+
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../../../source/core/BSP-Dungeon.hpp"
+
+/// @brief Brought over from DungeonOne for test case purposes
+/// @return vector of
```
karensuzue, line 7: You haven't finished this sentence

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,257 @@
+
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../../../source/core/BSP-Dungeon.hpp"
+
+/// @brief Brought over from DungeonOne for test case purposes
+/// @return vector of 
+inline static constexpr std::array<std::pair<std::size_t, double>, 6> ROOM_DATA{{
```
karensuzue, line 8: I think adding a comment here describing what this is would be helpful

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,257 @@
...
+		{6, 1},
+	}};
+
+inline static const std::string PREFIX = "one_pool/room_";
+
+static cse498::WeightedSet<std::string> MakeRoomPool() {
+	cse498::WeightedSet<std::string> rooms;
+	
+	for (const auto& [num, weight] : ROOM_DATA) {
+		auto result = rooms.Insert(PREFIX + std::to_string(num) + ".txt", weight);
```
karensuzue, line 23: Ah so that was a set of room IDs and weights

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -0,0 +1,257 @@
...
+        BSP.ClearState();
+
+        CHECK(BSP.GetBSPTree().size() == 0);
+        CHECK(BSP.GetLeafNodes().size() == 0);
+
+        BSP.RepopulateTree();
+
+        CHECK(BSP.GetBSPTree().size() != 0);
+        CHECK(BSP.GetLeafNodes().size() != 0);
+
```
karensuzue, line 93: To me, the name of this test case is a bit misleading, because you generate a full BSP tree before changing its settings (which doesn't update the tree). So you're not really reverting the object to a previous state. Also, I can't find any function named RegenerateObjectState(), I'm assuming that's legacy?

group_specific_content/Group-15/WorldGeneration.hpp
karensuzue, line 33: So this is the ASCII grid

group_specific_content/Group-15/WorldGeneration.hpp
```diff
@@ -39,26 +39,25 @@ namespace cse498 {
     public: 
 
         /// @brief Creates and initializes BSP Tree, RoomHolder, and grid for outputting dungeon level
-        WorldGen() 
+        WorldGeneration() 
             : mBSP(), //For now, the constructor for BSP_tree room creation is going to generate rooms immediately when initialized, will reformat as level specifications become more detailed
               mRoomHolder(), 
               mGrid(mBSP.GetHeight(), std::string(mBSP.GetWidth(), '#'))
```
karensuzue, line 45: The grid starts out full of solid walls…

group_specific_content/Group-15/WorldGeneration.hpp
karensuzue, line 115: I think some safeguards/checks might be needed here before indexing into room, just to make sure it's not empty

karensuzue, line 69: I would add another assert for room_width

source/tools/Random.hpp
```diff
@@ -49,193 +51,178 @@ namespace cse498 {
...
             // changes each time a number is generated
-            struct xoshiro256pp_state {
-                uint64_t s[STATE_NUMBER];
+            struct Xoshiro256ppState {
+                std::array<uint64_t, STATE_NUMBER> state;
             };
 
-            struct splitmix64_state {
-                uint64_t s;
+            struct Splitmix64State {
```
karensuzue, line 58: I think you could use type aliasing here instead of a struct with a single member

source/tools/Random.hpp
karensuzue, line 1: As far as I know, all the math here looks good! Good job!!

source/tools/Random.hpp
```diff
@@ -49,193 +51,178 @@ namespace cse498 {
...
-            int GetInt(int i_min = 0, int i_max = 100) {
-                // Error handling: i_min must be less than i_max
-                if (i_min > i_max) {
-                    throw std::runtime_error("cse498::Random::GetInt(): i_min must be less than i_max.");
+            /* @brief Templated function to generate and return values for integral values
+            * @param The range of values(inclusive) to generate a number between. min must be <= max.
+            * @return A random value in range of the specified type
+            */
+            template <std::integral T>
+            std::expected<T, std::string> GetValue(T min, T max) {
```
karensuzue, line 171: Good use of templates and concepts here, very well done

karensuzue (approved):

source/Worlds/Dungeon/DungeonBase.hpp
GumdropHorse: Fixed!

source/tools/WeightedSet.hpp
GumdropHorse: Fixed!

group_specific_content/Group-15/WorldGeneration.hpp
GumdropHorse: Fixed!

GumdropHorse: Fixed!

source/tools/Random.hpp
GumdropHorse: Fixed!

PR #70, opened 2026-03-25
Group18

Mehver:
For review, you can focus on WebLayout and WebCanvas. Also, there're some more advanced C++ features used in tools/Color.hpp

Read this document to see the specific methods for compiling and viewing the demo.
https://github.com/CSE498/Spring2026-CompanyA/blob/group18/main/source/Interfaces/WebUI/README.md

On the wiki there is the documentation for APIs:
https://github.com/CSE498/Spring2026-CompanyA/wiki/(Group18)-WebUI#group18-webui-module-apis


source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -124,112 +148,133 @@ void WebLayout::SetAlignment(IDomElement* elem, Alignment a) noexcept {
...
+  style.set("opacity", mOpacity.has_value() ? std::to_string(*mOpacity) : "");
+  style.set("boxShadow", mBoxShadow);
+  style.set("width", px(mWidth));
+  style.set("height", px(mHeight));
 }
 
-void WebLayout::ApplyLayout(val& style) noexcept {
+void WebLayout::ApplyLayout(val style) noexcept {
   // Helper lambda to convert Justification to CSS justify-content string.
   auto getJustifyStr = [](Justification j) -> std::string {
```
Cobathon: Very good use of lambdas, scoped correctly and is very easy to read

I would recommend adding constexpr like this
constexpr auto getJustifyStr...
will help with being compile time evaluable.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -245,6 +290,8 @@ void WebLayout::ApplyLayout(val& style) noexcept {
         return "space-around";
       case Justification::SpaceEvenly:
         return "space-evenly";
+      case Justification::None:
+        return "";
     }
     assert(false && "Unhandled Justification");
```
Cobathon, line 320: I recommend just adding at return ""; default case for saftey

source/Interfaces/WebUI/WebImage/WebImage.cpp
```diff
@@ -224,10 +266,14 @@ void WebImage::Resize(int width_px, int height_px, bool maintain_aspect_ratio) {
...
-    mElement["style"].set("objectFit",
-        std::string(maintain_aspect_ratio ? "contain" : "fill"));
+    val style = mElement["style"];
+    const auto set_style = [&](const char* property, const std::string& value) {
+      style.set(property, value);
+    };
+
+    set_style("width", ToPx(width_px));
+    set_style("height", ToPx(height_px));
+    set_style("objectFit", maintain_aspect_ratio ? "contain" : "fill");
```
Cobathon: You can use a template for this,

Something like this:
template <typename T, typename Func>
void Apply(emscripten::val& style, const char* key, const std::optional<T>& val, Func transform)
{
    if (val) {
        style.set(key, transform(*val));
    }
}

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -13,108 +26,119 @@ int WebLayout::mNextIdCounter = 1;
 /// Helper function to convert a pixel value to a CSS string.
 /// @param v Pixel value (non-negative)
 /// @return CSS string like "10px" or empty string if v < 0
-static std::string px(int v) noexcept {
-  if (v < 0) return std::string();
-  return std::to_string(v) + "px";
+static std::string px(std::optional<int> v) noexcept {
```
Cobathon: Changing px(int v) to -> px(std::optional<int> v) Isn't more beneficial than before as it will now force optional everywhere, instead I recommend using both, or just make it a template:

template<typename T>
std::string px(const std::optional<T>& v)

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include <sstream>
+#include <algorithm>
+
+using namespace cse498;
+
+namespace {
+constexpr cse498::Color kDimGray = cse498::Color::FromRGB255(0xcc, 0xcc, 0xcc);
+constexpr cse498::Color kCanvasBg = cse498::Color::FromRGB255(0x0c, 0x10, 0x17);
+static_assert(kDimGray.R() == 0xcc && kDimGray.G() == 0xcc && kDimGray.B() == 0xcc);
+static_assert(kCanvasBg.R() == 0x0c && kCanvasBg.G() == 0x10 && kCanvasBg.B() == 0x17);
```
Cobathon, line 23: Good use of static_assert to catch bad color definitions at compile time!

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+
+      mCanvas->Clear();
+
+      int top = topOffset + drawSize;
+      for (const auto & row : symbol_grid) {
+        int left = leftOffset + drawSize / 2;
+        for (char cell : row) {
+          mCanvas->DrawTexture(mTextures.at(cell).as_handle(), left, top, desiredScale);
```
Cobathon, line 208: throws if any grid character or agent symbol is missing from the map. An improvement could be asserting with a test that enumerates all symbols the world can emit

source/Group18_main.cpp
```diff
@@ -0,0 +1,89 @@
...
+  return EM_TRUE; 
+}
+};
+
+} // anonymous namespace
+
+/// @brief Main entry point for the application.
+/// @return Exit code.
+int main() {
+  static App * app = new App();
```
Cobathon: This statement leaks for the lifetime of the page, you can either change it to just static App app; or document that the leak is intentional.

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,220 @@
...
+      canvasHeight = canvasHeight / dpr;
+
+      const int totalPixelWidth = grid.GetWidth() * IMAGE_SIZE;
+      const int totalPixelHeight = grid.GetHeight() * IMAGE_SIZE;
+
+      const double widthRatio = static_cast<double>(canvasWidth) / totalPixelWidth;
+      const double heightRatio = static_cast<double>(canvasHeight) / totalPixelHeight;
+
+      const double desiredScale = widthRatio <= heightRatio ? widthRatio : heightRatio;
+      const int drawSize = gsl::narrow_cast<int>(IMAGE_SIZE * desiredScale);
```
Cobathon, line 184: This is good! This is being explicit about converting a float scale to an integer pixel size.

source/Interfaces/WebUI/interface/MockWorld.hpp
```diff
@@ -0,0 +1,151 @@
...
+  /// @return Reference to the added interface.
+  template <std::derived_from<InterfaceBase> T>
+  InterfaceBase & AddInterface(std::string interfaceName="None") {
+    mInterface = std::make_unique<T>(agent_set.size(), interfaceName, *this);
+    
+    InterfaceBase & interfaceRef = *mInterface;
+    ConfigAgent(*mInterface);
+
+    if (mInterface->Initialize() == false) {
+      std::cerr << "Failed to initialize interface" << std::endl;
```
Cobathon: If this fails only the cerr is used, the function still ends up returning a reference to the interface. The callers are not able to assert success without parsing the logs.

tests/Interfaces/WebUI/WebButtonTest.cpp
Cobathon: This might be a little confusing, cse498::WebLayout is already being used so this declaration on line 18 doesn't really do much if I'm reading into this right.

Cobathon (commented):
This is overall well-structured and fun to read into! The build worked fine for me, no issues there and the gameplay was really interesting to see. I love the progress that is being made and all the art is well put together.

 Interoperability & API
I see a lot of pieces that the other groups will be able to use, WebInterface fits the model by extending InterfaceBase and then the Group18_main.cpp shows how it runs with MockWorld and PacingAgent. The Readme is well put together as well

Although, some parts like asset paths and fetch assume a fixed layout, a few of the issues I brought up with some of my comments, such as mTextures.at which can throw if a symbol is missing. Also MockWorkd::Interface only prints if init fails instead of reporting failure through API.

Testing & error checking
One thing that I noticed is that there are no tests that drive WebInterface end-to-end with the world. fetch/image loading does not show clear checks for bad http responses or failed bitmap creation. And as stated, mTextures.at can crash the demo if there isn't a symbol with a texture.

Ebizi:
Hi! I was able to build the files and nothing obtusely stuck out as wrong to me as I ran through the gameplay, so congrats on that! As suggested, I looked at mostly WebLayout and WebCanvas alongside Color, and I had a brisker view of most of the other classes

Implementation Quality:
Overall the codebase looks really strong, there wasn't anything that immediately jumped out as inefficient or anything of the sort. I thought the Color class was actually implemented especially well, the validation of color constants at compile time was really nice.

Testing and Error Checking:
As with Cobathon, I also didn't see any tests WebInterface's end to end with the world. Another thing I didn't see (and it could just be that I completely missed it somewhere) is that I didn't see any unit tests or anything like that for WebCanvas, which I figured would have been important since it has some pretty meaningful behavior.

I was not able to see anything egregious to fix, but there were a few things I thought could use fixing (I could not get the code highlighting portion to work for the life of me so I apologize if this part is unclear:

WebLayout.cpp:

The Clear() function:

void WebLayout::Clear() noexcept {
  for (const auto& elem : mChildren) {
    elem->Unmount();
  }
  mChildren.clear();
  mParams.clear();
}

does not call SetParent(nullptr) to reset the child's parent pointer, so if a child checks for a parent after being cleared, it's gonna be holding a dangling pointer.

In ApplyLayout, there's already a lambda getAlignItemsStr. Then in ApplyChildren there's a separate inline switch doing the same thing:

switch (a) {
    case Alignment::Start:  alignStr = "flex-start"; break;
    case Alignment::Center: alignStr = "center";     break;

You could add [[nodiscard]] to AddElement and RemoveElement if you wanted, like this:

[[nodiscard]] bool AddElement(IDomElement* elem, Alignment align = Alignment::None) noexcept;
[[nodiscard]] bool RemoveElement(IDomElement* elem) noexcept;

Color looked well-implemented for the most part and I couldn't see any glaring issues with it when I looked over it, WebCanvas also looked fine although admittedly I am less well-versed in that sort of code. Overall, good work!

Ebizi (commented):
Hi! I was able to build the files and nothing obtusely stuck out as wrong to me as I ran through the gameplay, so congrats on that! As suggested, I looked at mostly WebLayout and WebCanvas alongside Color, and I had a brisker view of most of the other classes

Implementation Quality:
Overall the codebase looks really strong, there wasn't anything that immediately jumped out as inefficient or anything of the sort. I thought the Color class was actually implemented especially well, the validation of color constants at compile time was really nice.

Testing and Error Checking:
As with Cobathon, I also didn't see any tests WebInterface's end to end with the world. Another thing I didn't see (and it could just be that I completely missed it somewhere) is that I didn't see any unit tests or anything like that for WebCanvas, which I figured would have been important since it has some pretty meaningful behavior.

I was not able to see anything egregious to fix, but there were a few things I thought could use fixing (I could not get the code highlighting portion to work for the life of me so I apologize if this part is unclear:

WebLayout.cpp:

The Clear() function:

void WebLayout::Clear() noexcept {
  for (const auto& elem : mChildren) {
    elem->Unmount();
  }
  mChildren.clear();
  mParams.clear();
}

does not call SetParent(nullptr) to reset the child's parent pointer, so if a child checks for a parent after being cleared, it's gonna be holding a dangling pointer.

In ApplyLayout, there's already a lambda getAlignItemsStr. Then in ApplyChildren there's a separate inline switch doing the same thing:

switch (a) {
    case Alignment::Start:  alignStr = "flex-start"; break;
    case Alignment::Center: alignStr = "center";     break;

You could add [[nodiscard]] to AddElement and RemoveElement if you wanted, like this:

[[nodiscard]] bool AddElement(IDomElement* elem, Alignment align = Alignment::None) noexcept;
[[nodiscard]] bool RemoveElement(IDomElement* elem) noexcept;

Color looked well-implemented for the most part and I couldn't see any glaring issues with it when I looked over it, WebCanvas also looked fine although admittedly I am less well-versed in that sort of code. Overall, good work!

joeyhyun1:
Everything compiled and ran correctly and the main classes that I looked at were WebCanvas, Weblayout, and Color as well as looking the other classes. 

Code Design : 
Informative Identifiers - Good work here. Class names, member variables, and method names, are all clear. Variables are also named correctly. 
Avoids "magic numbers" - good work here. I did not see any examples of this.
Intuitive and functional Interface - Looks good, all member variables are private and the public ones are only for functionality which makes it easy to use.
Informative Comments - Files are well documented. The file-level comments explaining the purpose of each class is very helpful. Parameter documentation is consistent throughout the code. Comments explain why some coding decisions were made, and not just what the code does is great. 
Consistent Style - Excellent consistency across the files. Member variables use m prefix, Functions are PascalCase, Constants use k prefix, enums use Pascal case. 
Visual appearance - the code is well structured and easy to read through. Consistent indentation and use of whitespace is good. 

Interoperability and API : 
Interfaces for other groups - The classes are well-designed for cross-group integration. 
ICanvasElement : Clean, other groups only need to implement Draw(WebCanvas& canvas)
Color - also great for integration, any group can use it for color handling
Group18_main.cpp : Builds and provides a working example of the groups module. Demonstrates how WebUI works with the world systems. This demo shows good integration with other group's modules and demonstrated cross group integration. ReadMe is also well put together to make it easier for group to understand.

source/Interfaces/WebUI/interface/WebInterface.cpp
anaghaj9, line 1: Code Design: action_map is used throughout SelectAction(), but its definition isn't visible - it should be clearly documented as a member and ideally initialized in the constructor with all expected keys, since mTextures.at() will throw/crash if a key is missing.

Advanced C++: Excellent use of constexpr & static_assert for color validation at compile time! The use of std::ranged::for_each with a lambda for iterating cell types is clean and natural. gsl::narrow_cast for the IMAGE_SIZE * desiredScale conversion shows good awareness of narrowing pitfalls. The CellXToLeft and CellYToTop lambdas captured by reference, including locals - worth noting this is safe here since they don't escape scope, but being explicit about capture intent is good practice! Overall, great use of advanced C++ features!!

source/Interfaces/WebUI/WebImage/WebImage.cpp
anaghaj9, line 1: Code Design:

**Double callback bug** — HandleLoad() calls the callback twice: once with if (mOnLoadCallback) { mOnLoadCallback(); } and then again with InvokeIfSet(mOnLoadCallback). Same issue in HandleError(). This is a real correctness bug, not just a style issue

**Duplicate code throughout** — SyncFromModel() and ApplyPlaceholder() both have the old inline style-setting code AND the new lambda-based version side-by-side, meaning both run on every call
Constructor contradicts its own comment - the comment says "constructs a detached WebImage without mounting into document body" but the body calls doc["body"].call<void>("appendChild", mElement), doing the exact opposite

**Global mutable state** — sImageRegistry and sNextRegistryId are file-scope globals. A better design would encapsulate the registry as a static member of WebImage, making ownership and lifetime clearer

Advanced C++:

The apply_dimension and set_style lambdas are redefined from scratch inside multiple methods. These should be private helper methods or a template. to avoid repetition. Getter functions like Id() are good candidates for [[nodiscard]] since ignoring their return value is almost certainly a mistake

source/Interfaces/WebUI/WebLayout/WebLayout.hpp
anaghaj9, line 1: Code Design:

The header shows both old and new versions of member variables side by side (e.g., int mSpacing = 0 alongside std::optional<int> mSpacing{})  - this suggests an incomplete refactor and likely won't compile cleanly

ApplyStyling and ApplyLayout are each declared twice with different signatures (val& vs val) - one of these needs to be removed

Advanced C++:

Switching from raw int sentinel values (like -1) to std::optional<int> for styling properties is a good modern C++ improvement!

demos/WebUI/Group18_main.cpp
anaghaj9, line 1: Code Design:

**Intentional memory leak needs documentation**:  static App * app = new App() leaks for the lifetime of the page. This should either be changed to static App app; (no heap allocation needed) or have a comment explicitly stating the leak is intentional and explaining why. Leaving it ambiguous is a design smell

**if (!app) return EM_FALSE is a dead check**:  app is set from new App() two lines above and passed directly into the loop. It can never be null here, so this guard is misleading. If the intent is defensive programming, it should be documented; otherwise it should be removed

Wrapping everything in cse498 namespace is good practice:  it prevents RunningMockWorld and App from polluting the global namespace


Advanced C++

static_cast<App*>(userData) is the right cast to use here over C-style casting when converting from a void* - shows awareness of C++ cast semantics

delta >= 0 check is too loose - since double subtraction can produce very small negative values due to floating point precision, this check is reasonable, but it silently drops frames rather than clamping. A more robust approach would be **const double delta = std::max(0.0, currentTimeMs - app->last_time_ms)** and removing the conditional entirely

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
anaghaj9, line 1: Code Design:

Duplicate member variables: Both m_elements and mElements are used throughout the file (e.g., AddElement pushes to both, ClearElements clears both, RenderFrame iterates both). This is clearly an incomplete refactor and is a real bug,  elements would be stored and rendered twice

Inconsistent naming convention: The leftover m_elements (snake_case) alongside mElements (camelCase) breaks the naming convention used everywhere else in the codebase. The old name should be fully removed

Constructor is written twice: There's a one-liner constructor WebCanvas::WebCanvas(std::string id)  : mId(std::move(id)) immediately followed by the full implementation. The first version appears to be a leftover that was never removed

SyncFromModel calls RenderFrame():  The comment says "safe no-op placeholder" but then immediately calls RenderFrame(), which is the opposite of a no-op. The comment and the implementation contradict each other

DrawPoint hardcodes a magic color string: The old version passes "#000000" as the stroke color, while the new version correctly uses Color::FromRGB255(0,0,0).ToHex(). The old version should be fully removed since magic color strings are exactly what the Color class was built to replace
 

Advanced C++:

std::move(id) in the constructor to avoid copying the string is correct

Using std::stable_sort for z-index ordering is the right choice -  preserving insertion order for equal z-values is important for predictable rendering

ordered.reserve(mElements.size()) before the loop is good practice to avoid reallocations

Clear() uses cssColor == "" — this should be cssColor.empty() instead, which is the idiomatic and slightly more efficient C++ way to check for an empty string

anaghaj9 (commented):
I left comments on the files mentioned in the PR!

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
yhenry6, line 1: Implementation Quality
- No warnings during compilation
- Proper uses of namespace
- Adheres to the C++ standard: Nothing in this file clearly showed undefined behavior
- Appropriate Modularity: The code has useful helper functions like px, ApplyStyling, ApplyLayout, and ApplyChildren.
- Uses modern C++ constructs: Uses modern C++ constructs throughout the code, like std::optional and lambdas in the first helper function.
- Uses standard library algorithms: There were multiple good uses of std::find instead of creating manual loops.
- Pointer handling: It does use raw pointers for child elements, but is done carefully with new and delete.
- const correctness: No const member functions, but const parameters like in ElementID& id = elem->Id() were properly used. 

Advanced C++ Features
The code uses Advanced C++ features well. Like the usage of lambda in ApplyLayout, which kept enum-to-string conversion local and readable. It also used std::optional to represent style properties that may or may not be set, which is cleaner than using magic values.

source/tools/Color.hpp
yhenry6, line 1: Advanced C++ features: 
The code uses many of the Advanced C++ features. There were good uses of templates, lambdas, ranges, constexpr, and value semantics throughout the code. 

Implementation Quality: 
- No warnings during compilation
- Proper uses of namespace
- Adheres to the C++ standard: Nothing in this file clearly showed undefined behavior
- Appropriate Modularity: The code has several helper functions. 
- Uses modern C++ constructs: Good use of std::optional for parsing 
- Uses standard library algorithms: Uses std::find instead of manual loops, like in the ParseNamedColor function
- Pointer handling: Memory management is handled properly, and raw pointers aren't used. 
- Const correctness: Member functions throughout the code that do not modify state are properly marked const.

tests/Interfaces/WebUI/WebButtonTest.cpp
Pizarita: This is a stub so WebButton tests compile without depending on the real WebLayout. I will remove once WebLayout is ready to integrate :D

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: Thank you for the suggestion. The code has been updated (5e0054b3cf7abb71bd0a53535c3ff858dc6a9d39).

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: Thank you for the suggestion. The code has been updated (5e0054b3cf7abb71bd0a53535c3ff858dc6a9d39).

demos/WebUI/Group18_main.cpp
Mehver: Thank you for the suggestion. The code has been updated (df027eadb5d746edd2bb79d2a3af9749a053434a),

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Thank you for the suggestion. The code has been updated (ed5791f90db8107b149d30fe21c405bcf7be5e33).

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Thank you for the suggestion. The code has been updated (ed5791f90db8107b149d30fe21c405bcf7be5e33).

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Thank you for the suggestion. The code has been updated (ed5791f90db8107b149d30fe21c405bcf7be5e33).

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Thank you for the suggestion. The code has been updated (ed5791f90db8107b149d30fe21c405bcf7be5e33).

source/Group18_main.cpp
Mehver: Thank you for the suggestion. The code has been updated (df027eadb5d746edd2bb79d2a3af9749a053434a),

source/Interfaces/WebUI/WebLayout/WebLayout.hpp
Mehver: Thank you for your suggestions. However, for the old and new variables issue you mentioned, I was unable to locate it. My guess is that during your review process, the interface you were using or the fed into the LLM may have included Git’s historical diffs, which led to this misunderstanding.

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
Plutolinus: Thank you for your suggestions！

source/Interfaces/WebUI/WebImage/WebImage.cpp
Plutolinus: Thanks, these are good catches.

tests/Interfaces/WebUI/WebButtonTest.cpp
motowsk2: Removed stub with WebLayout in after adding some new webutton tests

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: Thank you for the suggestion. The code has been updated (a7953c7359d11101dc8915178e417755c97df7a9).

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
```diff
@@ -22,60 +35,92 @@ extern "C" {
...
 // WebCanvas
 // --------------------
 
-WebCanvas::WebCanvas(std::string id)
-  : mId(std::move(id))
+/// @brief Constructs a WebCanvas with the given canvas element id.
+/// @param id DOM id of the \<canvas\> element; defaults to "web-canvas" if empty.
+WebCanvas::WebCanvas(const std::string & id)
 {
+    mId = std::move(id);
```
karensuzue, line 58: Since `id` is **const**, casting it to an rvalue reference (from lvalue reference) doesn't give you any performance boost. it's still going to call `std::string`'s copy assignment operator. I'd change this to `mId = id`, since that is effectively what it already does.

You could technically do this:
`WebCanvas(std::string id) { mId = std::move(id); }` 

but I don't really see the benefit. Here you're doing a copy (from pass by value) AND a move (by calling `std::string`'s copy assignment operator).

Also, I would recommend using the initializer list instead of assigning to member variables inside the constructor body, I believe that is more computationally efficient.

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
```diff
@@ -22,60 +35,92 @@ extern "C" {
...
     if (mId.empty()) {
         mId = "web-canvas";
     }
+#ifdef __EMSCRIPTEN__
+    mElement = GetDocument().call<emscripten::val>("getElementById", mId);
+    mExisting = true;
```
karensuzue, line 64: I think this may be misleading, because I doubt that an element with `mId` has been created at that point. So `mExisting` is technically false... unless the element is first created separately in the DOM and this class is then attached to it.

I would suggest doing something similar to what you have in the demo for this class:
```
let c = document.getElementById("web-canvas");
        if (!c) {
            c = document.createElement("canvas");
            c.id = "web-canvas";
            document.body.appendChild(c);
        }
```

source/Interfaces/WebUI/WebCanvas/WebCanvasDemo.cpp
karensuzue, line 1: I tried compiling this file (following your instructions), and it seems to be generating some kind of type‑mismatch error
```
 error: no viable conversion from 'const std::string_view'
      (aka 'const basic_string_view<char>') to 'const std::string' (aka 'const basic_string<char>')
   89 |     g_canvas->Clear(kBgColor);
```
The functions in your WebCanvas class take an std::string as an argument, not an std::string_view. The error went away when I reverted std::string_view back to std::string :)

The second set of errors I found were related to the DrawText function, which currently requires 6 arguments instead of 5. Once I added the additional argument (fontFamily), the error went away.

source/Interfaces/WebUI/WebCanvas/WebCanvasDemo.cpp
```diff
@@ -1,74 +1,146 @@
...
+/// @brief Entry point: constructs real WebImage, WebButton, WebTextbox objects,
+///        registers them with the canvas, and starts the animation loop.
 int main() {
     static cse498::WebCanvas canvas("web-canvas");
     g_canvas = &canvas;
 
-    // Create a <canvas> if not present, and set a fixed size & basic styles.
     EM_ASM({
         let c = document.getElementById("web-canvas");
         if (!c) {
```
karensuzue, line 151: Yeah, this is related to what I said earlier in WebCanvas.cpp about the constructor not actually creating a new element with the given id (in this case "web-canvas"). It is likely calling the code inside this block under `if (!c)` here.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -14,107 +27,142 @@ int WebLayout::mNextIdCounter = 1;
...
+  }
+  assert(false && "Unhandled Alignment");
+  return "";
+};
+
+/// @brief Constructs a WebLayout, creating or adopting a root DOM element.
+/// @param rootId Optional ID; if an element with this ID exists in the DOM it
+///               is adopted as root, otherwise a new <div> is created and
+///               appended to <body\>.
 WebLayout::WebLayout(const std::string& rootId) noexcept {
```
karensuzue, line 64: I see some potential issues with the current constructor implementation. The first is in the else block on lines 68-75:
```
    if (!found.isNull() && !found.isUndefined()) {
      mElement = found;
      mExisting = true;
    } else {
      // create a container div with the requested id and append to body
      mElement = mDocument.call<val>("createElement", std::string("div"));
      mElement.set("id", rootId);
    }
```
If the element with the given ID is not found, we create it. In that case, I think it would make sense to mark `mExisting` as `true` there as well.

Another issue is that all of this only happens when `rootId` is not empty. You said it's optional, so what happens if it is empty? If you don't want to handle that case, I would suggest making `rootId` required instead :)

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -14,107 +27,142 @@ int WebLayout::mNextIdCounter = 1;
...
+/// @brief Sets the cross-axis alignment for all child elements.
+/// @param a Desired Alignment value.
 void WebLayout::SetAlignItems(Alignment a) noexcept { mAlignItems = a; }
 
-/// Sets the gap/spacing between child elements.
+/// @brief Sets the gap/spacing between child elements.
+/// @param pxv Spacing in pixels; must be >= 0.
 void WebLayout::SetSpacing(int pxv) noexcept {
-  assert(pxv >= 0);
+  if (pxv < 0) return;
```
karensuzue, line 115: I vaguely remember saying to avoid using assert for input validation, but I still think some form of error handling would be beneficial here. I would use std::expected, but returning a bool, like you did in AddElement, could also work...

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -14,107 +27,142 @@ int WebLayout::mNextIdCounter = 1;
...
-  }
+  // remove DOM element from this layout
+  IDomElement::RemoveChild(elem);
   return true;
 }
 
+/// @brief Updates the per-child alignment stored in this layout.
+/// @param elem Pointer to the child element to update.
+/// @param a    New Alignment value for that child.
 void WebLayout::SetAlignment(IDomElement* elem, Alignment a) noexcept {
```
karensuzue, line 166: Similarly, I think returning a bool here or using std::expected would be a good idea, since there is a failure condition :)

source/Interfaces/WebUI/internal/IDomElement.hpp
```diff
@@ -76,16 +142,21 @@ class IDomElement {
...
   void ClearFreePosition() { mTop = mLeft = -1; }
 
- private:
+ protected:
+  emscripten::val mElement = emscripten::val::undefined();  ///< HTML element for this dom element
+  IDomElement * mParent = nullptr;                          ///< DOM element for the parent of this element
+  std::string mId{};                                        ///< DOM Id of this element
+  bool mExisting{false};                                    ///< True if this element is just hooking into an existing element
+
   int mGridRow = -1;  ///< Grid row position (-1 = unset)
```
karensuzue, line 160: Now that I think about it, this could be the perfect opportunity to use std::optional

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
karensuzue, line 385: I think this could be cleaner and more readable if you used std::optional for the grid coordinates in IDomElement :)

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -364,18 +407,24 @@ void WebLayout::ApplyChildren() noexcept {
   }
 }
 
+/// @brief Applies all pending layout and style changes to the DOM.
+/// Calls ApplyStyling(), ApplyLayout(), and ApplyChildren() in order.
+/// No-op when the layout is currently hidden (mIsVisible == false).
```
karensuzue, line 412: Just in case, have you considered the scenario where the layout was previously visible, and then made hidden? I assume ApplyStyling handles this...

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
```diff
@@ -364,18 +407,24 @@ void WebLayout::ApplyChildren() noexcept {
...
+  if (!mIsVisible) return;
   ApplyChildren();
 }
 
+/// @brief Unmounts and removes all child elements from this layout.
 void WebLayout::Clear() noexcept {
   // remove children from DOM by unmounting them
   for (const auto& elem : mChildren) {
-    elem->Unmount();
+    if (elem) elem->Unmount();
```
karensuzue, line 427: There is an interesting call chain here. From what I can tell, this calls IDomElement's Unmount, which then calls WebLayout's RemoveChild, which calls WebLayout's RemoveElement, which then calls IDomElement's RemoveChild. I would recommend simplifying or collapsing this flow if possible.

tests/Interfaces/WebUI/WebCanvasTest.cpp
karensuzue, line 1: I think some of my comments from the last round of grading about the misuse of REQUIRE_NOTHROW and verifying at the DOM level still have not been addressed here, but I do see that you have added a few more good tests :)

tests/Interfaces/WebUI/WebCanvasTest.cpp
```diff
@@ -300,4 +350,223 @@ TEST_CASE("WebCanvas immediate-mode primitives are safe to call (expanded)", "[w
...
+    SpyElement* raw1 = e1.get();
+    SpyElement* raw2 = e2.get();
+
+    canvas.AddElement(std::move(e1));
+    canvas.AddElement(std::move(e2));
+
+    canvas.RenderFrame();
+    CHECK(raw1->drawCount == 1);
+    CHECK(raw2->drawCount == 1);
+
```
karensuzue, line 488: I would also check to see if the vector 'order' here

tests/Interfaces/WebUI/WebLayoutTest.cpp
karensuzue, line 1: You've done a great job improving your test cases and adding DOM-level verification - just out of curiosity, why the change from TEST_CASE to TEST_CASE_METHOD? :)

tests/Interfaces/WebUI/WebLayoutTest.cpp
```diff
@@ -474,287 +650,525 @@ TEST_CASE("WebLayout complex styling scenario", "[weblayout][styling][integratio
...
+    val parent2 = GetElement(layout2.Id());
+
+    CHECK(GetProperty(parent1, "children")["length"].as<int>() == 1);
+    CHECK(GetProperty(parent2, "children")["length"].as<int>() == 0);
+
+    elem.MountToLayout(layout2);
+    layout1.Apply();
+    layout2.Apply();
+
+    CHECK(GetProperty(parent2, "children")["length"].as<int>() == 1);
```
karensuzue, line 1100: I would also check for parent1 here

source/tools/Color.hpp
karensuzue, line 1: Good work! One thing to note is that the file is missing documentation, and adding it would benefit the project greatly - for example, what is ExpandNibble? It took me a moment to figure out the purpose of this function :)

tests/tools/ColorTest.cpp
karensuzue, line 1: Good tests!

source/Interfaces/WebUI/interface/MockWorld.hpp
```diff
@@ -0,0 +1,175 @@
...
+    MOVE_DOWN,
+    MOVE_LEFT,
+    MOVE_RIGHT,
+    INTERACT,
+    QUIT
+  };
+
+protected:
+  /// @brief Array of floor cell type IDs for easy access.
+  std::array<size_t, 3> mFloorIds;
```
karensuzue: I'm guessing there are three different floor cell types (and 5 impassable cell types)?

source/Interfaces/WebUI/interface/MockWorld.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+  /// @brief Array of impassable cell type IDs for easy access.
+  std::array<size_t, 5> mImpassableIds;
+
+  /// @brief Unique pointer to the interface used by this world.
+  std::unique_ptr<InterfaceBase> mInterface = nullptr;
```
karensuzue, line 41: I think having the world own the player agent is a good decision

source/Interfaces/WebUI/interface/MockWorld.hpp
```diff
@@ -0,0 +1,175 @@
...
+    mInterface->SetActionResult(result);
+    if (!mInterface->IsPaused()) {
+      RunAgents();
+    }
+    mInterface->RenderFrame();
+  }
+
+  /// @brief Cleans up and exits the application.
+  virtual void Teardown() {
+    mInterface.release();
```
karensuzue, line 170: I think this is going to leak memory, since unique_ptr's `release()` does not delete the object. It doesn't matter much here because we exit right away (in fact, I think simply calling `exit(0)` would suffice) but it would be better to use `.reset()`.

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,261 @@
...
+  mElements.emplace_back(std::make_unique<WebTextbox>("press Escape to resume, or Q to quit."));
+  WebTextbox* descPtr = static_cast<WebTextbox*>(mElements.back().get());
+  descPtr->SetAlignment(WebTextbox::TextAlign::Center);
+  descPtr->SetFontSize(24.0f);
+  descPtr->SetColor(kDimGray.ToHex());
+  descPtr->MountToLayout(*mPauseMenu);
+
+  auto cellTypes = world.GetGrid().GetCellTypes();
+
+  std::ranges::for_each(cellTypes, [this](const CellType & cell){
```
karensuzue, line 110: I'm pretty sure a range-based for loop would be better here :)

source/Interfaces/WebUI/InputManager.hpp
```diff
@@ -0,0 +1,71 @@
...
+  ~InputManager();
+    
+  /// @enum ActiveAction
+  /// @brief Enumeration of possible user actions.
+  enum class ActiveAction {Up, Left, Down, Right, Interact, Pause, Quit, None};
+
+  /// @brief Callback for key down events.
+  /// @param eventType The event type.
+  /// @param keyEvent The keyboard event.
+  /// @param userData User data pointer.
```
karensuzue, line 30: What exactly is "user data"? You should specify the expected object type here. It seems like you're expecting an InputManager‑derived class (since you static-cast it in the function definition), or maybe some other Emscripten specific thing. Either way, more documentation is needed :)

source/Interfaces/WebUI/InputManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+InputManager::~InputManager() {
+  emscripten_html5_remove_all_event_listeners();
+}
+
+EM_BOOL InputManager::OnKeyDown(int eventType, const EmscriptenKeyboardEvent * keyEvent, void * userData) {
+  if (eventType != EMSCRIPTEN_EVENT_KEYDOWN || userData == nullptr || keyEvent->repeat) return EM_FALSE;
+
+  auto manager = static_cast<InputManager *>(userData);
```
karensuzue, line 35: Okay, this is scaring me a little bit (the fact that you have to take in a void* and then static_cast it back) but I'm assuming it's some kind of Emscripten workaround. At least you're using static_cast so it should be safe :)

source/Interfaces/WebUI/InputManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+  std::string lowerKey = ToLower(key);
+
+  // special case pause button
+  if (lowerKey == "escape") {
+    manager->mInterface.HandlePause();
+    return EM_TRUE;
+  }
+
+  if (!manager->IsKeyPressed()) manager->SetTapAction(lowerKey);
+  manager->AddKeyPressed(lowerKey);
```
karensuzue, line 48: It would be great if you could add some documentation explaining the overall idea of what you're doing and why you're adding pressed keys into the queue, etc.

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,261 @@
...
+}
+
+bool WebInterface::Initialize() {
+  std::array<const char *, 6> actions{
+    "up",
+    "left",
+    "down",
+    "right",
+    "interact",
+    "quit"
```
karensuzue, line 156: This works, but I would prefer not to hard-code these action names here. It would be better to centralize them so they are shared across methods/classes instead of duplicated in multiple places.

Also, I noticed that "pause" is not included in this list. Make sure that is intentional and not an oversight :)

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,261 @@
...
+      const int topOffset = (canvasHeight - drawSize * grid.GetHeight()) / 2;
+
+      auto CellXToLeft = [&leftOffset, &drawSize](auto cellX){ return (leftOffset + drawSize * (cellX + 1)) - drawSize / 2; };
+      auto CellYToTop = [&topOffset, &drawSize](auto cellY){ return topOffset + drawSize * (cellY + 1); };
+
+      std::vector<std::string> symbolGrid(grid.GetHeight());
+
+      // Load the world into the symbolGrid;
+      for (size_t y=0; y < grid.GetHeight(); ++y) {
+        symbolGrid[y].resize(grid.GetWidth());
```
karensuzue, line 196: Resizing here might be costly? I would do this:
```
std::vector<std::string> symbolGrid(grid.GetHeight(), std::string(grid.GetWidth(), '\0'));
for (size_t y = 0; y < grid.GetHeight(); ++y) {
  for (size_t x = 0; x < grid.GetWidth(); ++x) {
    symbolGrid[y][x] = grid.GetSymbol(WorldPosition{x, y});
  }
}
```
I'm not sure how much of an improvement that makes, but it does read a bit cleaner

source/Interfaces/WebUI/interface/WebInterface.cpp
```diff
@@ -0,0 +1,261 @@
...
+  if (mPaused) {
+    if (!mPauseMenu->IsVisible()) mPauseMenu->ToggleVisibility();
+  } else {
+    if (mPauseMenu->IsVisible()) mPauseMenu->ToggleVisibility();
+
+    auto itemIds = world.GetKnownItems(*this);
+    auto agentIds = world.GetKnownAgents(*this);
+    auto grid = world.GetGrid();
+
+    DrawGrid(grid, itemIds, agentIds);
```
karensuzue, line 249: It seems like your code redraws entire frames rather than only the tiles that changed, and for a project like this that's totally fine! 
I do wonder if there are some small optimizations worth considering. For example, every time you call DrawGrid, you reconstruct a symbolGrid. If you could initialize it once and update it only when needed, that would be great. The function also recalculates things like canvas height and width, which could be computed once and cached inside the class. If you have time, I'd definitely recommend making these adjustments.

source/Interfaces/WebUI/interface/MockWorld.hpp
```diff
@@ -0,0 +1,175 @@
...
+    });
+  }
+
+  ~MockWorld() = default;
+
+  /// @brief Executes an action for the given agent.
+  /// @param agent The agent performing the action.
+  /// @param action_id The ID of the action to perform.
+  /// @return True if the action was successful, false otherwise.
+  virtual int DoAction(AgentBase& agent, size_t action_id) override {
```
karensuzue, line 128: Keep in mind that you're returning boolean values, but your function signature expects ints. It'll still work but is misleading

karensuzue (approved):

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
Mehver: Thanks, fixed in fcf33258e1f80d67d985fe2bac54d530ae3e7715.

source/Interfaces/WebUI/WebCanvas/WebCanvas.cpp
Mehver: Good point. In this UI the canvas is expected to already exist in the DOM, since this class is intended to bind to the single pre-created canvas rather than create one. I updated the logic so mExisting reflects the actual lookup result, and report an error when the element is missing instead of marking it as existing unconditionally. Ref: 85c682f8aaf7b7f708cd8637b4f70766a2a968c6

source/Interfaces/WebUI/WebCanvas/WebCanvasDemo.cpp
Mehver: Solved. Ref: 
https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3055077154
https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3083210519

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Good catch! After tracing usage, I believe this variable does not affect the current program logic path (@alexelkins23 correct me if I was wrong).

The main issue here is semantic clarity: mExisting was meant to indicate “adopted from a pre-existing DOM element,” not general element validity, so the current name is misleading.

Since we’re close to completion, I’d rather avoid changing runtime ownership behavior now and keep this as a safe cleanup only: clarify the comment and rename the variable.

Fix ref: f6cd1b53400d3a2b7f49f175a4b98b7ef03285dd 0cf78dccc76ed2ef5dc59a42655477dda129f2af

Now it looks like:

https://github.com/CSE498/Spring2026-CompanyA/blob/0cf78dccc76ed2ef5dc59a42655477dda129f2af/source/Interfaces/WebUI/internal/IDomElement.hpp#L166-L170

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: I think that in Emscripten, the top priority should be to avoid exceptions, while assertions have a relatively smaller impact (and also perhaps we can disable it with a complie parameter `-s ASSERTIONS=0` in the final stage). Moreover, regardless of that, there seems like still some codes in World or Agent are using exceptions. I suspect this may not actually lead to any noticeable performance bottleneck; and even if it does, we can address it before the final release by performing a repository-wide, cross-module cleanup of error handling. For example, we could provide a unified utility class to enable switching between different error reporting methods, such as browser alert pop-ups, the console, or other approaches, and then refactor the entire codebase to route all error handling through this utility. This would likely be the most practical solution.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: Thank you for the suggestion. It would indeed be better to add checks here. However, changing it to `[[nodiscard]] bool` would require relatively significant modifications. Additionally, introducing C++ exceptions would incur considerable runtime performance overhead. Therefore, I chose to add error logging in place, printing messages to the browser console.

Fix ref: 35de223a50a1187850a829ad4392e3c895b1508d

https://github.com/CSE498/Spring2026-CompanyA/blob/35de223a50a1187850a829ad4392e3c895b1508d/source/Interfaces/WebUI/WebLayout/WebLayout.cpp#L169-L189

source/Interfaces/WebUI/internal/IDomElement.hpp
Mehver: Thanks for your advice. This could ref to the answer in https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3106581845

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: > Another issue is that all of this only happens when rootId is not empty. You said it's optional, so what happens if it is empty? If you don't want to handle that case, I would suggest making rootId required instead :)

Good question. I didn’t write this piece of code, so I also spent quite a while thinking about it when I first read it. Then I realized that although it isn’t written as an explicit `if-else`, the early `return` makes the control flow effectively equivalent. So the empty `rootId` case was already handled. To improve readability, I reorganized the structure.

Fix ref: 2aeb6a7d7994eddf5355cb6acced9db49e0fb106

https://github.com/CSE498/Spring2026-CompanyA/blob/2aeb6a7d7994eddf5355cb6acced9db49e0fb106/source/Interfaces/WebUI/WebLayout/WebLayout.cpp#L62-L90

tests/Interfaces/WebUI/WebLayoutTest.cpp
Mehver: For `TEST_CASE_METHOD`, I assume (I’m not the author) that it provides a shared context compared to `TEST_CASE` (that `SharedWebContext` in this case) which can be shared across tests.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: That’s a great suggestion. However, this is already a foundational implementation, and as an interface definition, many classes inherit from it. Given that there’s only about a week left in the project timeline, I think we should stick with the existing implementation for now.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: @alexelkins23 could you check this? I got completely no idea for it... Is this feature even still using?

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
Mehver: I will try to refactor it. The functions currently in the call chain are:

https://github.com/CSE498/Spring2026-CompanyA/blob/2aeb6a7d7994eddf5355cb6acced9db49e0fb106/source/Interfaces/WebUI/WebLayout/WebLayout.cpp#L433-L442

https://github.com/CSE498/Spring2026-CompanyA/blob/c9e01f93d5a8fae83882e49e6a03e3d84821855a/source/Interfaces/WebUI/internal/IDomElement.hpp#L45-L54

https://github.com/CSE498/Spring2026-CompanyA/blob/c9e01f93d5a8fae83882e49e6a03e3d84821855a/source/Interfaces/WebUI/WebLayout/WebLayout.cpp#L282-L285

https://github.com/CSE498/Spring2026-CompanyA/blob/c9e01f93d5a8fae83882e49e6a03e3d84821855a/source/Interfaces/WebUI/WebLayout/WebLayout.cpp#L152-L167

https://github.com/CSE498/Spring2026-CompanyA/blob/c9e01f93d5a8fae83882e49e6a03e3d84821855a/source/Interfaces/WebUI/internal/IDomElement.hpp#L77-L116

Mehver: With the analysis helped by LLM, aside from the overly long call chain, there are currently two main sources of UB:

1. **`Clear()` is currently unsafe.** While iterating over `mChildren` using a range-based for loop, it calls `elem->Unmount()` on each element. However, `Unmount()` in turn calls back into the parent layout via `RemoveChild(this)`, which then proceeds into `RemoveElement()` and erases the current element from `mChildren`. In other words, the code is modifying the same `vector` while iterating over it.

2. **`AddElement()` sets `elem->SetParent(this)`, but `RemoveElement()` does not clear it.** `RemoveElement()` only removes the element from `mChildren` / `mParams`, and then calls `IDomElement::RemoveChild(elem)` on the DOM side, but it does not reset the child’s parent pointer. This results in inconsistency between the "model state" and the "DOM state."

Mehver: 56fd11af68d854d25865c0bc274203f57cfc57a3 I made an in-place change using a lambda and added an `@todo` comment to track the call chain updates. As for the other functions in the call chain, I’m concerned there might be callers that weren’t picked up by the IDE’s analysis (though that’s unlikely), so I’ve left them in place for now.

Mehver: It seems that the convoluted implementation in the past was a compromise caused by TDD-driven development. With my current optimized implementation, I’m now encountering test failures.

```bash
this.program is a Catch v2.13.10 host application.
Run with -? for options

-------------------------------------------------------------------------------
WebLayout Clear removes all elements
-------------------------------------------------------------------------------
tests/Interfaces/WebUI/WebLayoutTest.cpp:364
...............................................................................

tests/Interfaces/WebUI/WebLayoutTest.cpp:380: FAILED:
  CHECK( elem1.unmountCount == 1 )
with expansion:
  0 == 1

tests/Interfaces/WebUI/WebLayoutTest.cpp:381: FAILED:
  CHECK( elem2.unmountCount == 1 )
with expansion:
  0 == 1

===============================================================================
test cases:  58 | 57 passed | 1 failed
assertions: 208 | 206 passed | 2 failed
```

---

@alexelkins23 Can I modify the tests? Doing so would change the semantics of `Clear`, and I’m not sure whether that might have ripple effects on other parts of the code.

---

@karensuzue I’d prefer to avoid overly aggressive refactoring, so I may not be able to significantly improve the semantics. If I want to avoid undefined behavior without modifying the tests, I might implement it like this. Do you think this would be an acceptable fix?

```c++
void WebLayout::Clear() noexcept {
    while (!mChildren.empty()) {
        IDomElement* element = mChildren.back();
        if (!element) {
            mChildren.pop_back();
            continue;
        }
        element->Unmount();
    }
    mParams.clear();
}
```

tests/Interfaces/WebUI/WebCanvasTest.cpp
Mehver: fceb337e75b8cdc3d9b056bf01a20d9804f6581b Optimized the implementation and scope of the tests. And since verifying the DOM using Catch2 in the Emscripten C++ runtime is inherently complex, the actual rendering behavior and 2D rendering checks have been handled by `WebCanvasDemo`.

tests/Interfaces/WebUI/WebCanvasTest.cpp
Mehver: Done fea7e927b485bc85f04e32e3bc152b08dcc24cee

tests/Interfaces/WebUI/WebLayoutTest.cpp
Mehver: Done ad1bab15103c4da8affbcbdf97410db8e0a6c24c

source/tools/Color.hpp
Mehver: Done 58667f5d14b5bb08c5a3c7209675fc1c8bbff7c8

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: @alexelkins23 This seems to be related to your current work?

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: Very good suggestion! However, `Teardown()` no longer exists, so I will mark this issue as resolved.

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: This issue has been fixed already. Previously, we used `for_each` mainly to satisfy the requirement of using advanced C++ features :D

source/Interfaces/WebUI/InputManager.hpp
Mehver: @alexelkins23 @motowsk2 @hoangtu0701 Is this part something one of you is still developing as a work in progress? I think it would be worthwhile to clearly define the parameter types once the API is finalized.

source/Interfaces/WebUI/InputManager.cpp
Mehver: I will consolidate this issue with [https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3066533357](https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3066533357) and close this tracker.

source/Interfaces/WebUI/InputManager.hpp
Mehver: https://github.com/CSE498/Spring2026-CompanyA/pull/70#discussion_r3066564997

source/Interfaces/WebUI/InputManager.cpp
Mehver: @alexelkins23 Perhaps you could add docs for it quickly?

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: Thanks for your advice! But the relevant code has been refactored.

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: Thanks for your advice! But the relevant code has been refactored.

source/Interfaces/WebUI/interface/WebInterface.cpp
Mehver: > It seems like your code redraws entire frames rather than only the tiles that changed, and for a project like this that's totally fine!

Yeah true lol.

> I do wonder if there are some small optimizations worth considering. For example, every time you call DrawGrid, you reconstruct a symbolGrid. If you could initialize it once and update it only when needed, that would be great. The function also recalculates things like canvas height and width, which could be computed once and cached inside the class. If you have time, I'd definitely recommend making these adjustments.

This is definitely a very insightful idea, thanks for sharing!

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: I think it would be best to modify the function signature directly here. However, this change may affect callers (such as `WebInterface`). @alexelkins23 Let me know once your current work is finished, and I’ll refactor this part accordingly. Or you can fix it while you’re at it if you’d like.

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: I think `MockWorld.hpp` is about to be deprecated, so it might be acceptable to ignore this change.

source/Interfaces/WebUI/interface/MockWorld.hpp
Mehver: I think `MockWorld.hpp` is about to be deprecated, so it might be acceptable to ignore this change.

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
karensuzue: I think that's fine :) I just wanted to point it out because you're more likely to run into issues with such a long call chain. If you can fix it that would be great, but I agree we don't really have time for a full overhaul

motowsk2: Added the new clear function after talking with our group during last meeting. Test suite and demo are all working after change. Resolving converstaion

source/Interfaces/WebUI/WebLayout/WebLayout.cpp
alexelkins23: If I'm understanding correctly, yes ApplyStyling() updates the visibility and it is always called when Apply() is called, so the layout should always reflect mIsVisible after you call Apply(). I removed the No-op line in the comment as it is no longer accurate.

source/Interfaces/WebUI/InputManager.hpp
alexelkins23: 24e971c0861e9684d9b82060b8fb97406ef8734b I updated this documentation to be more descriptive. Basically emscripten requires callbacks to be static functions so you need to pass a pointer to any data you want access to inside the callback. The emscripten callback signature also requires it to be a void * so you can pass anything you want.

source/Interfaces/WebUI/InputManager.cpp
alexelkins23: 24e971c0861e9684d9b82060b8fb97406ef8734b I added more descriptive documentation here. We use a deque to support multiple keys being pressed at once. The most recently pressed key will be at the front and used for input, then if they release that key, the next most recent key they pressed will be at the front and used for input.

PR #72, opened 2026-03-26
Group 2 Module and 2 Classes (read desc)

ZodiacalBee:
The two updated classes are 
PathGenerator & ActionMap

Look at all change files but the most important module changes are in 
AgentBase
WorldBase agent additions
and everything under Agents/


source/core/WorldPosition.hpp
```diff
@@ -61,7 +58,7 @@ namespace cse498 {
     auto operator<=>(const WorldPosition &) const = default;
```
hoangtu0701: You default operator<=> but then provide a custom operator== that uses
epsilon comparison. This means == and <=> can disagree - two positions
could be "equal" by == but not by <=>. This can cause subtle bugs in
ordered containers. Either make both exact or both epsilon-based.

source/tools/DamageCalculator.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "CombatStats.hpp"
+
+namespace cse498 {
+
+  class DamageCalculator {
+  public:
+    [[nodiscard]] static double Calculate(const CombatStats &attacker, const CombatStats &defender);
```
hoangtu0701: Calculate could be constexpr since it only does basic math with
std::max. Would let it be evaluated at compile time if needed.

tests/tools/TestPathVector.cpp
```diff
@@ -0,0 +1,259 @@
...
+ * @file TestPathVector.cpp
+ * @author Logan Rimarcik
+ */
+
+
+
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "tools/PathVector.hpp"
+#include "core/WorldPosition.hpp"
+#include <iostream>
```
hoangtu0701: iostream is included but never used

tests/Agents/TestSkeleton.cpp
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "../../source/Worlds/MazeWorld.hpp"
+
+using cse498::WorldGrid;
+using cse498::MazeWorld;
+
+TEST_CASE("Skeleton Tests Cumulative -- Categorize later", "[Skeleton Basics]")
```
hoangtu0701: This only has one test that checks GetNumAgents() >= 1. Needs actual
tests for MazeWorld and skeleton behaviors.

tests/Agents/TestPlayer.cpp
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+}
+
+
+
+using cse498::PlayerAgent;
+
+
+TEST_CASE("Main Player", "[Inventory]")
```
hoangtu0701, line 34: Only one test here that checks the player exists and hand is null. Should test more of PlayerAgent's API like health, damage, actions, etc.

tests/tools/TestMemoFunction.cpp
```diff
@@ -70,63 +70,64 @@ TEST_CASE("MemoFunction: Class member functions w/ multiple inputs and types", "
...
+//    REQUIRE_THROWS_AS(mf6(), std::logic_error);
+//    REQUIRE(mf6.CacheSize() == 0);
+
+//    std::function<int()> F;
+//    cse498::MemoFunction<int> mf7(F);
+//    REQUIRE_THROWS_AS(mf7(), std::bad_function_call);
+//   REQUIRE(mf7.CacheSize() == 0);
+//}
+
+//int Fib(int x) {
```
hoangtu0701, line 98: Big chunk of commented-out test code here. Either remove it or convert it to actual tests

tests/tools/TestPathGenerator.cpp
```diff
@@ -383,7 +405,7 @@ TEST_CASE("Circular Path Generation -- Simple Cases", "[CirclePath]")
     CHECK(WorldPathApprox(WorldPath(circ2_exp_path_to), circ2_exp.value().mPathToCircle));
 
     cse498::CircleWorld2 circle_world2;
-    PathRequest circle_request2({}, cse498::AgentAbility(), circle_world2.GetGrid());
+    PathRequest circle_request2({}, circle_world2.GetGrid());
 
     // Generation with center on wall
     // FOREWARNING: Looking at this in desmos can be confusing it can look like "Why did it take that path??"
```
hoangtu0701, line 411: Three empty test cases in this file (circle edge cases, rectangle edge cases, IsPathClear). I think they should be filled in or removed (empty TEST_CASEs give a false sense of coverage)

tests/Agents/TestPacingAgent.cpp
```diff
@@ -0,0 +1,141 @@
...
+  auto agent_ptr = std::make_unique<PacingAgent>(1, "pacer", world);
+
+  // WorldBase::AddAgent will call ConfigAgent and then Initialize().
+  auto &pacer = world.AddAgent(std::move(agent_ptr));
+
+  // Initialize should have succeeded because ConfigAgent added all four actions.
+  CHECK(pacer.Initialize());
+}
+
+TEST_CASE("PacingAgent vertical pacing toggles on failure", "[PacingAgent][Vertical]") {
```
hoangtu0701, line 88: Good tests for wall bounce in both directions. You could also test what happens if the agent starts on a wall tile or at the grid edge.

tests/Agents/TestTrading.cpp
```diff
@@ -0,0 +1,168 @@
...
+    REQUIRE(enemy.GetGoldDrop() == 9);
+
+    REQUIRE(enemy.ClaimGoldDrop() == 9);
+    REQUIRE(enemy.ClaimGoldDrop() == 0);
+
+    enemy.ResetGoldDropClaim();
+    REQUIRE(enemy.ClaimGoldDrop() == 9);
+}
+
+TEST_CASE("Buying a limited merchant item reduces stock and player gold", "[agents][trade][buy]")
```
hoangtu0701, line 97: Solid trade tests here covering buying, selling new items, and restocking. But I see no test for buying when the player doesn't have enough gold, or buying more than the merchant has in stock.

hoangtu0701 (commented):
Implementation Quality:
The new files are clean overall. DamageCalculator is simple and correct, CombatStats is minimal and fine. Minor things like a dead include in TestPathVector and DamageCalculator::Calculate could be constexpr.

Testing & Error Checking:
Strong test suites for ActionMap, BehaviorTree, WorldPath, PathVector, PathGenerator, Trading, Inventory, PacingAgent, and FarmingAgent. PathGenerator testing is particularly really good with hand-verified test worlds :D

Main gaps are TestSkeleton and TestPlayer, which are basically placeholders with
one test each. Several empty test cases in PathGenerator and a block of commented-out code in MemoFunction tests should be cleaned up. Trading tests could cover failure cases (insufficient gold, out of stock). No tests exist for DamageCalculator or CombatStats.

source/Agents/PlayerFeatures/Inventory.hpp
```diff
@@ -0,0 +1,315 @@
...
+namespace cse498
+{
+/**
+ * This is a temporary placeholding struct that will be deleted once dependancy of Item class is met
+ * NOTE!!!
+ * NOTE!!!
+ * This is NOT my class. This is a placeholder from Group 15 that I needed to implement inventory
+ * I wish an early pull request was done so we could both access this, but this class is weird.
+ */
+class Item // I am not responsible for tests!!
```
hoangtu0701: This placeholder Item class lives inside Inventory.hpp but it's a full class with virtual methods that other files depend on. Should be in its own header. Also the "I am not responsible for tests!!" comment should be removed before final submission.

source/Agents/PlayerFeatures/Inventory.hpp
```diff
@@ -0,0 +1,315 @@
...
+ *
+ * NOTE: 2 types of items:
+ * Unique items max stack size == 1, size_t id identifies them as different objs
+ * Non-Unique items max stack size is MAX_STACK_SIZE, size_t id is IGNORED, std::string name
+ * identifies them as different objects otherwise they stack.
+ * ^^ if you want two variants of the same item with the same name, then add a std::string id as well
+ * I was unable to push hard enough to convince everyone for this change.
+ */
+
+#ifndef SPRING2026COMPANYA_INVENTORY_H
```
hoangtu0701: Rest of the codebase uses #pragma once but this file uses a traditional include guard. Same in Enemy.hpp, AgentFactory.hpp, MovementTypes.hpp, FarmingAgent.hpp, PlayerAgent.hpp. We should have one style across the group.

source/Agents/AgentFactory.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+        if (!enemy)
+            return Node::Status::Failure;
+
+        // No-op attack: enemy will stand still once in range.
+        return Node::Status::Success;
+    }));
+
+    // Chase action: when not in range, move one step toward the player using PathGenerator.
+    auto chasePlayer = TreeBuilder::Act("Chase Player", [enemy, &world](ExecutionContext& ctx)
```
hoangtu0701: This lambda captures &world by reference and enemy by raw pointer. If the world or enemy gets destroyed while this tree still exists, you get a dangling reference. The skeleton tree has the same issue. Maybe document the lifetime requirement or use a safer pattern.

source/Agents/Classic/MovementTypes.hpp
```diff
@@ -0,0 +1,30 @@
...
+#ifndef SPRING2026COMPANYA_MOVEMENTTYPES_H
+#define SPRING2026COMPANYA_MOVEMENTTYPES_H
+#include <cstddef>
+
+namespace cse498
+{
+
+    class MovementTypes
+    {
+    private:
```
hoangtu0701, line 24: Empty private section so just remove it. Also this class is only static methods with no state. Could just be free functions in a namespace instead of a class, or delete the constructor to prevent instantiation

source/Agents/MovementTypes.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "Worlds/DemoSimpleWorldG2Actions.hpp"
+
+namespace G2 = cse498::DemoSimpleWorldG2Actions;
+
+namespace cse498
+{
+
+std::size_t MovementTypes::DeltaToMoveAction(double dx, double dy) {
+    if (dx == 0.0 && dy == -1.0) {
```
hoangtu0701: Comparing doubles with == is fragile. If dx or dy come from floating point math and aren't exactly 1.0 or -1.0, none of these conditions will match and you silently get REMAIN_STILL. Maybe try using an epsilon comparison or comparing signs instead

source/Agents/PlayerAgent.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <iostream>
+
+namespace cse498 {
+
+PlayerAgent::PlayerAgent(size_t id, const std::string &name, const WorldBase &world)
+    : InterfaceBase(id, name, world) {}
+
+size_t PlayerAgent::SelectAction(const WorldGrid & /*grid*/) {
+    char input = '\0';
+    std::cout << "WASD move, E interact with NPC/enemy, X quit.\n> ";
```
hoangtu0701: SelectAction does std::cin/std::cout directly. This works for the terminal demo but will likely block if used with a GUI or web interface

source/Agents/Enemy.hpp
```diff
@@ -0,0 +1,56 @@
...
+#ifndef CSE498COMPA_ENEMY_H
+#define CSE498COMPA_ENEMY_H
+#include "core/AgentBase.hpp"
+
+namespace cse498
+{
+    class Enemy : public AgentBase
+    {
+    private:
+        double mAttackRange = 3; // 3 block units
```
hoangtu0701: No setter for mAttackRange. If different enemy types need different ranges (which AgentFactory suggests), there's no way to configure it from outside

tests/Agents/TestInventory.cpp
```diff
@@ -0,0 +1,367 @@
...
+    // some edge cases:
+    auto result = inv.RemoveItem("none", 3);
+    CHECK(result == 3);
+
+
+    // Should hit assert:
+    auto item = std::make_unique<RealUItem>(1, 5, world);
+    CHECK(item->IsUnique());
+
+    // inv.AddItem(std::move(item), 5);
```
hoangtu0701, line 387: If this was supposed to test the assert for adding multiple unique items, consider using
REQUIRE_THROWS or just remove it

tests/Agents/TestInventory.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+TEST_CASE("Swap Slots simple", "[inventory, swap]")
+{
+    // Swapping null item and real item
+    Inventory inv;
+    CHECK(inv.GetHandSlotIndex() == 0);
+    CHECK(inv.GetHand() == nullptr);
+    inv.AddItem(std::make_unique<MockTestItem>(0, "TestItem"), 3);
+
+    // std::cout << "Inventory before swap";
```
hoangtu0701: Couple of commented-out cout lines in this test to be cleaned up

hoangtu0701 (commented):
Additional review comments focused on the new module code under Agents/:
 
Implementation Quality:
TradeSystem has solid error handling with proper rollback on failed transactions. Good class hierarchy with MerchantAgent -> FarmingAgent. Main concerns are: placeholder Item class embedded in Inventory.hpp should be its own file, AgentFactory captures raw pointers and references in lambdas with no lifetime safety, MovementTypes compares doubles with == which is fragile, and Enemy has no setter for attack range despite the factory pattern suggesting different enemy types.
 
Testing & Error Checking:
Inventory tests are thorough - good coverage of slots, stacking, unique vs non-unique, overflow/underflow, and swap. Trading tests cover the happy paths well but are missing failure cases (insufficient gold, out of stock). Some commented-out test code should also be cleaned up.

source/core/WorldBase.hpp
```diff
@@ -122,6 +160,19 @@ namespace cse498 {
       }
     }
 
+    /// Remove agents that are dead: call OnDestroy() then erase from agent_set.
+    virtual void RemoveDeadAgents() {
```
ZhixiangMiao, line 241: Interoperability & API: When you remove the dead agent, I also recommend removing its id. After you remove an agent, positions inside the agent_set shift, but agent ids do not. This may cause the id to be assigned to the wrong agent. I recommend using a map or pair instead of a set to map the agent with its id. When the agent is dead, remove its id as well.

AB123456kkkkk (commented):
Advanced C++ features.
ActionMap -- Template function that validates a trigger request, looks up action, checks expected signature, and packs arguments into vector<any> was added. Very helpful function. Also used lambdas. 

Inventory.hpp — Constexpr vairables such as BACKPACK_SIZE, HOTBAR_SIZE, and INVENTORY_SIZE. Template function is added.

PathGenerator.hpp — Added constexpr variables such as STEP_SIZE, MAX_SEARCH_DISTANCE. Also added a constexpr function EuclidanDistance. Aded a tempalte function.

MerchantAgent.cpp -- Lambda is used.

Code Design 
PathGenerator.cpp — One thing that I liked well in this file is has lots of informative comments. An exmample of this is in the function PathGenerator::CircleTravel PathGenerator::IsTravelableCircle, the if (!pth.empty()) and then there is a bunch of comments explaining what is being chcked.  “it brings us to the right tile but we change the actual position on that tile to what we want.” “This will be generally better though some spots that are longer but if we can get to the tile then we can get to a specific position on that tile type of thing.” Throughout the file, there are many comments, and all of them are very helpful. PathGenerator.hpp — The file is very easy to scan and the code is visually well structured with good use of whitespace, indentation, and line breaks.
The rest of the files are very easy to read and have consistent naming conventions.

source/Agents/Classic/PlayerFeatures/Inventory.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (it == mItemMap.end() || it->second.empty())
+    {
+        return nullptr;
+    }
+
+    const size_t index = *it->second.begin();
+    return mInventory.at(index).GetItem();
+}
+
+size_t Inventory::RemoveUniqueItem(const size_t itemId)
```
ZhixiangMiao, line 130: Interoperability & API: The RemoveUniqueItem() function only resets the item slot, but it does not remove its index from mItemMap. It could affect the program when adding an item later on, such as when you remove and re-add the same item, it will only increase the quantity in your mItemMap without an actual item pointer. And also, I think this return should be void with no return type since it removes a unique item. If you really want to check if the unique item was removed you can return this function as a bool.

tests/Agents/TestFarmingAgent.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+TEST_CASE("FarmingAgent SelectAction returns 0", "[FarmingAgent]")
+{
+    TestWorld world;
+    FarmingAgent agent(2, "Farmer", world);
+
+    WorldGrid grid;
+    REQUIRE(agent.SelectAction(grid) == 0);
+}
+
```
ZhixiangMiao, line 66: Better add some comments for test cases to help reviewers understand better.

tests/Agents/TestInventory.cpp
```diff
@@ -0,0 +1,367 @@
...
+    inv.SwapSlots(0, Inventory::HOTBAR_SIZE);
+    // std::cout << "Inventory after swap";
+    // std::cout << inv;
+
+    CHECK(inv.GetHandSlotIndex() == 0);
+    CHECK(inv.GetHand()->GetName() == "TestItem");
+}
+
+
+TEST_CASE("Inventory Add and Remove", "[add, remove, inventory]")
```
ZhixiangMiao, line 223: Testing & Error Checking: Try testing an edge case, remove and then re-add the same item.

source/Agents/PlayerFeatures/Inventory.hpp
ZodiacalBee: interesting point. I never really look because it isn't actual code. I wonder how picky developers are about this

source/Agents/Classic/AgentFactory.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    auto patrol = std::make_unique<Enemy>(world.GetNextAgentId(), "Patrol", world);
+    patrol->SetLocation(spawn);
+    patrol->AddAction("left", 1).AddAction("right", 2);
+    patrol->SetBehaviorTree(CreatePatrolTree(patrol.get()));
+    return patrol;
+}
+
+
+std::unique_ptr<Node> AgentFactory::CreateEnemyFollowPlayerTree(Enemy *enemy,
```
ZhixiangMiao, line 190: Error Checking: Better add some conditions to check if the chased target is valid or not after world state changes, when an agent is dead and removed.

source/Agents/Classic/PlayerFeatures/Inventory.cpp
ZodiacalBee: fixed for the first part and no for the second. Returning as a bool would be confusing since the others all return as size_t which is important feedback to know how much of an item is removed or added. If you want bool checking then just do the operation and add == 0. (3 characters and now you have the same bool). It is consistent in this manner. Breaking consistency would be confusing

source/Agents/AgentFactory.cpp
ZodiacalBee: ?? enemy owns the tree by unique pointer. Enemy can't be nullptr here since it is literally made and passed into this function and can't ever be null in my mind since if it is then nothing owns the tree so the tree dies?

source/Agents/AgentFactory.cpp
```diff
@@ -0,0 +1,234 @@
...
+                ctx.mBlackboard.Set<std::size_t>("selected_action", DemoAct::REMAIN_STILL);
+                return Node::Status::Failure;
+            }
+            const WorldPosition epos = enemy->GetLocation().AsWorldPosition();
+            const WorldPosition ppos = world.GetAgent(targetAgentIndex).GetLocation().AsWorldPosition();
+            if (IsAdjacentForCombat(epos, ppos)) {
+                ctx.mBlackboard.Set<std::size_t>("selected_action", DemoAct::INTERACT);
+                return Node::Status::Success;
+            }
+            const WorldPosition next = PathGenerator::NextCardinalToward(epos, ppos);
```
ZhixiangMiao: Error Checking: Add IsWalkable() and FindShortestPath() in there to make sure the enemy walked on the walkable path and the shortest path to follow the player, and avoid bad moves. You used NextCardinalToward(), which can be directly turned into action. Try to use the similar logic as shown in CreateSkeletonTree() in there.

tests/tools/TestBehaviorTree.cpp
ZhixiangMiao, line 1: Better add some comments to help the reviewer understand better.

tests/Agents/TestSkeleton.cpp
ZhixiangMiao, line 1: Test & Error Checking: More tests should be added there, such as to test if a path is walkable, and logic when an enemy follows and chases the player. And these can help later when you fix your code, when the enemy gets stuck in obstacles.

source/core/WorldBase.hpp
```diff
@@ -91,17 +115,31 @@ namespace cse498 {
...
+    }
+
+    /**
+     * Other build system for agents. Agents that are not defined as classes, create the agent from the factory
+     * and pass it into here to add the agent.
+     * @param agent
+     * @return
+     */
+    template<typename AGENT_T>
+    AGENT_T & AddAgent(std::unique_ptr<AGENT_T> agent)
```
ZhixiangMiao, line 206: Interoperability & API: I suggest that you should fully establish the world before adding any agent, since the agent might be established in an awkward spot while making the world. To be more consistent, fully make the world first, then add agents in a derived class.

source/Agents/PlayerFeatures/Inventory.hpp
ZodiacalBee: changed

source/core/WorldBase.hpp
```diff
@@ -91,17 +115,31 @@ namespace cse498 {
     /// @return A reference to the newly created agent
     template <typename AGENT_T>
```
MouseCR, line 193: Perfect use of templates here making it possible to add any agent

source/tools/PathGenerator.cpp
```diff
@@ -403,7 +403,7 @@ std::optional<WorldPath> PathGenerator::FindManhattanPath(const WorldPosition &s
     auto fup_down = [&](const WorldPosition &start_pos)
```
MouseCR: Good use of templates here, makes the code concise and easy to understand

source/tools/WorldPath.cpp
MouseCR, line 1: This file uses lambdas and ranges well. They help keep the code compact, readable, and easier to follow without adding unnecessary code. These new topics are used in a meaningful ways.

source/Agents/Classic/PlayerFeatures/Inventory.cpp
```diff
@@ -0,0 +1,208 @@
+/**
+ * @file Inventory.cpp
+ * @author Logan Rimarcik
+ */
+
+#include "Inventory.hpp"
+
+#include <numeric>
+#include <ranges>
+
```
MouseCR, line 10: missing #include ostream

source/tools/PathGenerator.hpp
```diff
@@ -94,16 +92,17 @@ class PathGenerator
...
         /// This is for reconstruction. Not sure what is best here maybe std::optional<WorldPosition> and just copy store it
-        std::shared_ptr<ANode> mPrev;
-        ANode(const WorldPosition& pos, double g, double f, std::shared_ptr<ANode> other) : mPos(pos), mg(g), mf(f), mPrev(std::move(other)) {}
+        /// This is for ANode.
+        ANode* mPrev = nullptr;
+        ANode(const WorldPosition& pos, double g, double f, ANode* prev) : mPos(pos), mg(g), mf(f), mPrev(prev) {}
     };
 
     /**
      * comparison operaor for the priority queue
```
marksandychez: Minor spelling error in operator

marksandychez (commented):
### Interoperability and API Review

- There is a well formed and detailed wiki page dedicated to explaining major classes such as inventory, agents, and the demo world. This is likely the current best documentation in the company and is definitely style to copy for other groups moving forward. The inline code snippets are also very helpful for more key spots.

### PathGenerator:

- Header is well formed with meaningful comments in Doxygen styling
- Clear naming convention that follows expect company function names


### ActionMap:

- There are continued detailed function comments for all major functions along with detailed explanations and notes for implementations uses.
- Clear naming convention that follows expect company function names


### Additional Comments needed

- Some fairly simple files could also use some Doxygen comments for consistency between files, examples would be DamageCalculator and CombatStats. Not critical as they are so short but are technically less documented than others.


### General Notes: 
There are errors as a result of the cse498::Random when building, not related to the group's classes. These same errors prevent me from running the entire test suite. Cannot confirm the behavior of the tests, but based on formatting and review they seem to be meaninful assertions in a clear format. 

Example Driver produces a runnable executable that shows basic behavior built off of the simple view. There are meaningful outputs in each "turn" to describe the activities of the module.

source/tools/PathGenerator.cpp
MouseCR: Could be upgraded to a range::transform

source/tools/PathGenerator.cpp
```diff
@@ -337,22 +337,22 @@ std::optional<CirclePath> PathGenerator::FindRectangularLoopPath(const WorldPosi
     auto start_path = AStarSearch(agent_pos, request, MultiGoalHeuristic{goals});
     if (start_path.empty())
         return {};
-    assert(goals.contains(round(start_path.back())));
+    assert(goals.contains(Round(start_path.back())));
 
     const auto loop_start = start_path.back();
     std::vector<WorldPosition> loop_path;
     size_t start_index = 0;
     for (size_t i = 0; i < loop_vec.size(); i++)
```
MouseCR: could be upgraded to a find_if

source/Agents/Classic/PlayerFeatures/Inventory.hpp
```diff
@@ -0,0 +1,262 @@
...
+    Item* GetHand() const { return mInventory.at(mCurrentHotbarSlot).GetItem(); }
+    size_t GetHandQuantity() const { return mInventory.at(mCurrentHotbarSlot).GetQuantity(); }
+    [[nodiscard]] size_t GetHandSlotIndex() const { return mCurrentHotbarSlot; }
+
+    /**
+     * This object actually can become quite complex, and I don't want
+     * meddling with internal variables since it can break things
+     * This class SHOULD NOT BE USED OUTSIDE THIS CLASS but is public for testing
+     */
+    class InventorySlot
```
marksandychez, line 118: I understand the need for more than a forward declaration for this subclass, but InventorySlot is quite a substantial class in it's own right and could probably be moved to it's own header file and added to the #includes

source/Agents/MerchantAgent.hpp
```diff
@@ -0,0 +1,56 @@
...
+        [[nodiscard]] const TradeOffer* FindOffer(const std::string& itemName) const;
+
+        [[nodiscard]] std::size_t GetGold() const { return mGold; }
+        void AddGold(std::size_t amount) { mGold += amount; }
+        bool SpendGold(std::size_t amount);
+
+        TradeResult BuyFromMerchant(PlayerAgent& player, const std::string& itemName, std::size_t quantity = 1);
+        TradeResult SellToMerchant(PlayerAgent& player, const std::string& itemName, std::size_t quantity = 1);
+
+        [[nodiscard]] virtual std::string GetTradeGreeting() const;
```
marksandychez: Maybe allow these to be a mutable member variable and not const? Depending on merchant implementation the UI teams may need to differentiate the merchants by what they say and not how they look depending on how many sprites are available.

source/Agents/MovementTypes.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "MovementTypes.hpp"
+
+#include "Worlds/DemoSimpleWorldG2Actions.hpp"
+
+namespace G2 = cse498::DemoSimpleWorldG2Actions;
+
+namespace cse498
+{
+
+std::size_t MovementTypes::DeltaToMoveAction(double dx, double dy) {
```
marksandychez: I don't think there should be any doubles for movement, ideally everything is handled in tile math and then drawn based on the integer tiles moved

source/tools/PathGenerator.cpp
MouseCR, line 1: Overall the implementation is very good. A couple of loops here could be upgraded to be a bit more modern. Otherwise it compiles on mac/windows without errors and memory leaks.

source/core/AgentBase.hpp
```diff
@@ -7,36 +7,105 @@
...
-    int action_result = 1;
-
-    char symbol = '*';
+    int mActionResult = 1;
+
+    // TODO: I would guess this will become some struct with a URI reference to image & other metadata
+    char mSymbol = '*';
+
+    /// Health and alive state for takeDamage / isAlive / onDeath
+    double mHealth = 100.0;
```
marksandychez: Health can be done as a double, but integer math is much easier to display. The UI will likely truncate to an integer and may cause the infamous "200 pump that is actually 199.5" from Fortnite. Not critical, just something to note. As long as it's consistent between all agents, thats all that matters

source/tools/ActionMap/ActionMap.hpp
```diff
@@ -30,43 +19,55 @@
...
-         * Kept as a struct instead of mapping directly to std::function to allow
-         * for storing metadata (just a description for now maybe more later)
+         * Stored as a struct rather than mapping directly to std::function so the
+         * action can also carry metadata such as an optional description.
          */
         struct ActionEntry
         {
             std::type_index signature{typeid(void)}; // exact std::function signature we registered with
-            std::function<std::any(const std::vector<std::any>&)> invoker; // type-erased call wrapper
+            std::function<std::expected<std::any, InvokeError>(const std::vector<std::any>&)> invoker; // type-erased call wrapper
```
marksandychez: Leading m for the struct variables needed here

source/tools/DamageCalculator.hpp
marksandychez: Depending on how the player interacts, this might not be constexpr. I wouldn't say it's a guarantee to be constexpr based on user interaction

marksandychez (commented):
### Code Design Review


### Informative Identifiers

- Largely follows the leading m for member variables standards, comment noted for possible additions

### Avoids "Magic Numbers"

- When possible, numbers are set as constexpr correctly to be calculated at compile time and referenced as a variable as opposed to inline magic numbers

### Intuitive and Functional Interface

- Game demo explains the basic functionality and produces output to describe game state

### Informative Comments

- Where reasonable, there are helpful comments to explain logic and usage in files. 
- Issues to avoid and implementation notes included

### Consistent Style

- Style is consistent throughout files. Follows company standards.
- Variables are named accordingly

### Visual Appearance

- White spaced is used and standardized
- No line is too egregiously long


General Notes:
Repeat of the API review section, I cannot confirm the test suite as it errors when running due to the random tests. Not technically the groups code causing the issue, but cant confirm new tests.

ZhixiangMiao:
**Overall Summary:** 

**Interoperability and API:** When you remove something, make sure you fully remove it. For example, when you remove a dead agent, please also remove its id to avoid wrong id assignment after positions in agent_set shift after you removed an agent. 

**Testing and Error Checking:** In your CreateEnemyFollowPlayerTree() function, you should add some conditions to check if the chased target is valid or not after world state changes, when an agent is dead and removed. And you used NextCardinalToward(), which can be directly turned into action. Try to use the similar logic as shown in CreateSkeletonTree, like IsWalkable() and FindShortestPath() to make sure the enemy walks on the walkable path and the shortest path to follow the player, and avoid bad moves. More tests should be added into TestSkeleton.cpp, such as to test if a path is walkable, and logic when an enemy follows and chases the player. And these can help later when you fix your code, when the enemy gets stuck in obstacles.

source/Agents/Classic/PlayerFeatures/Inventory.cpp
ZodiacalBee: it is in header

source/Agents/Classic/PlayerFeatures/Inventory.hpp
ZodiacalBee: it could but it is also public and fully tested. I think putting it in this namespace as a class makes a lot of sense so I left it here since Inventory::InventorySlot is clear to me vs InventorySlot is not since Group14 also has an inventory. I could still move it out and put it in a namespace but I don't see the advantage

source/core/AgentBase.hpp
ZodiacalBee: agreed. Thanks for pointing it out. I'll update the wiki

tests/tools/TestPathGenerator.cpp
ZodiacalBee: fixed

tests/tools/TestPathVector.cpp
ZodiacalBee: fixed

source/tools/PathGenerator.cpp
ZodiacalBee: ok.

source/tools/ActionMap/ActionMap.hpp
Cobathon: Thank you!

source/tools/PathGenerator.cpp
ZodiacalBee: I redid this function

source/tools/PathGenerator.cpp
ZodiacalBee: redid this function

source/tools/DamageCalculator.hpp
ZodiacalBee: It can be constexpr based on the math and if it is capable it should be unless of course debugging issues or other things. We wouldn't want this function holding things back. Fixed.

tests/Agents/TestTrading.cpp
tmaks2: Thanks for catching this. Added some more tests.

source/tools/PathGenerator.hpp
tmaks2: Fixed

source/Agents/Classic/AgentFactory.cpp
ZodiacalBee: ok

source/Agents/AgentFactory.cpp
ZodiacalBee: This is for demo only. We wanted something simple with less possible stuff. Skeleton is worked out and has a proper test file with another sort of demo just not visual. Could do replacement but we wanted something working fast so this was done first. It doesn't mean this agent should last to the final build, it is a demo agent tree.

source/Agents/MerchantAgent.hpp
ZodiacalBee: changed.

source/Agents/MovementTypes.cpp
ZodiacalBee: ok

source/Agents/Classic/MovementTypes.hpp
ZodiacalBee: good point. Could just do these in a namespace, but I like things statically bound to a class. It feels more flexible plus this file changed quite a bit as is.

source/core/WorldBase.hpp
ZodiacalBee: I assume you are referring to the player who gets created and added early. I do recognize this as you need to call configAgent again on player once the derived config is defined and set position and other things but you would have to do that anyway. Initialization could cause issues if defined differently..? 

Worst case we move player to be just owned by unique_ptr and not in the vector of agents at all. We put it in the vector of agents because that is how it was defined in the example world. We are fine just excluding it if that is fine with everyone else. I'll leave it for now as it shouldn't cause issues... let us know if it does.

source/core/WorldBase.hpp
ZodiacalBee: 1. we changed it to an incrementing index instead of .size() so that if agents are removed and added then multiple duplicate ids are not given out. Id is unique until size_t overflows 
2. I want to change this since it would be faster lookup and I also want agent searching by tags and such or names efficiently, but we didn't want to creep too hard into this class because we already felt guilty about what we did. Hence we asked the world groups to do some things but everyone knows communication is bad and who knows what will happen.

tests/Agents/TestInventory.cpp
ZodiacalBee: oh I guess.  I left that intentionally if anyone wanted to see what it looks like if it was a confusing because inventory starts at HOTBAR_SIZE and the test itself seems kind of weird to me. I can remove it though..

tests/Agents/TestInventory.cpp
ZodiacalBee: require throws is for exceptions no..? we don't do those. It hits an assert and I wanted to check that. I know it doesn't run every time but having it to check easily is nice for me. I don't see why the comment hurts anyone. Is there some defined reason I should remove it?

tests/Agents/TestInventory.cpp
ZodiacalBee: I'll do that in the way that makes sense to me..

Mehver:
Hi, sorry for the late message. During Group 1’s review, I noticed these issues in `AgentAbility.hpp`. However, they mentioned that this file is maintained by your group, and I assume they just pulled your code. So I’m forwarding these to you here.

### 🟦 Consistent style

> [!NOTE]
>
> It is recommended to use `#pragma once` here.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/29431ff907303d62e1002c454ee06e1f03f71bab/source/tools/AgentAbility.hpp#L10-L11
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/29431ff907303d62e1002c454ee06e1f03f71bab/source/tools/AgentAbility.hpp#L23

### 🟪 Consistent style

> [!IMPORTANT]
>
> The variable here should follow the `mValue` naming convention.
>
> https://github.com/CSE498/Spring2026-CompanyA/blob/29431ff907303d62e1002c454ee06e1f03f71bab/source/tools/AgentAbility.hpp#L15-L17

---

Ref:

https://github.com/CSE498/Spring2026-CompanyA/pull/75#discussion_r3041168576

https://github.com/CSE498/Spring2026-CompanyA/pull/75#discussion_r3041173781

source/Agents/MovementTypes.cpp
Cobathon: Thanks for the suggestion. I think its best to keep MovementTypes on double + Approx for now so it stays aligned with PathVector / WorldPosition without a wider refactor.

source/Agents/Enemy.hpp
Cobathon: Attack range is AgentStats::mRange, we configure it through AgentBase::SetStats when the factory builds each enemy type, so per-type ranges are already set from outside. We don’t have a lone SetAttackRange on Enemy because it isn’t stored on an individual Enemy

ZodiacalBee:
> me they just pulled your code. So I’m forwarding these to yo

if you looked at our PR you would've seen that class was deleted so this was not needed.

demos/ClassicAgents/Group02_main.cpp
karensuzue, line 1: Great demo!

source/Worlds/DemoG2/DemoSimpleWorldG2.cpp
karensuzue, line 1: I think you did a great job putting all of this together. The logic makes a lot of sense and is easy to follow :)

source/core/WorldBase.hpp
karensuzue, line 1: I think these are good additions to the WorldBase class. My only "critique" is that you should probably update your documentation to follow the company standards :)

source/core/WorldBase.hpp
```diff
@@ -91,17 +187,31 @@ namespace cse498 {
...
     AGENT_T & AddAgent(std::string agent_name="None") {
```
karensuzue, line 194: I think the AddAgent(std::string) overload should also constrain AGENT_T to derive from AgentBase, similar to the one below it. I also think that a `requires std::derived_from<AGENT_T, AgentBase>` clause would probably be cleaner than a static_assert, so something like this:

```
template <typename AGENT_T>
requires std::derived_from<AGENT_T, AgentBase>
AGENT_T & AddAgent(std::string agent_name = "None") {
....
```

source/tools/PathGenerator.hpp
karensuzue, line 1: Overall this is a very thoroughly implemented class! Here are some suggestions:
1. I think some of the function definitions could be moved to the .cpp file, since the header is getting a little long and disorganized. 
2. AStarSearch is a templated function that takes a Heu object. I assume this is intended to be one of the two heuristic structures? If so, I think adding a requires clause/using concepts would make that clearer. So maybe something like this:
```
template <typename Heu>
concept PathHeuristic =
requires(const Heu& h, const WorldPosition& p) {
    { h(p) } -> std::convertible_to<double>;
    { h.equals(p) } -> std::convertible_to<bool>;
};

template <PathHeuristic Heu>
static std::vector<WorldPosition> AStarSearch(
...
```
3. I know A* is a well-known algorithm, but I think adding a few more comments in AStarSearch to describe what you're doing at each step could help break up the code and make it more readable.
4. Another reminder to update your documentation to match the current company standard :)

karensuzue, line 72: I think some of these variables could benefit from a comment explaining what they're meant for (like "EP" here, for example). It may become clearer once you start digging around, but adding comments makes things easier for anyone using your tool :)

source/tools/PathGenerator.hpp
```diff
@@ -379,43 +394,53 @@ template<typename Heu>
...
+
     const auto directions = std::to_array<PathVector>({
         {STEP_SIZE, 0}, {0, STEP_SIZE}, {-STEP_SIZE, 0}, {0, -STEP_SIZE},
         {STEP_SIZE, STEP_SIZE}, {-STEP_SIZE, STEP_SIZE}, {STEP_SIZE, -STEP_SIZE}, {-STEP_SIZE, -STEP_SIZE}
     });
-    std::priority_queue<std::shared_ptr<ANode>, std::vector<std::shared_ptr<ANode> >, ANodeCompare> pq;
+    ANode* closestNode = nullptr;
+
+    // pq can't own these because you can't std::move pq.top() and don't get it from pq.pop
+    std::vector<std::unique_ptr<ANode>> storage;
```
karensuzue, line 415: So storage owns all nodes, and the priority queue only points at them, that's a good idea

source/tools/ActionMap.hpp
```diff
@@ -30,43 +19,55 @@
...
+         * Stored as a struct rather than mapping directly to std::function so the
+         * action can also carry metadata such as an optional description.
          */
         struct ActionEntry
         {
-            std::type_index signature{typeid(void)}; // exact std::function signature we registered with
-            std::function<std::any(const std::vector<std::any>&)> invoker; // type-erased call wrapper
-            std::optional<std::string> description; // optional UI tooltip text
+            std::type_index mSignature{typeid(void)}; // exact std::function signature we registered with
+            std::function<std::expected<std::any, InvokeError>(const std::vector<std::any>&)> mInvoker; // type-erased call wrapper
```
karensuzue, line 45: I'm glad you considered adding std::expected here, and also creating an Error enum class :)

source/tools/ActionMap.hpp
```diff
@@ -78,47 +79,94 @@ namespace cse498
             return mActions.find(std::string(name));
```
karensuzue, line 79: I see why this conversion is here since the map stores std::string keys, but it does mean every lookup creates a temporary string. So while it works, it kind of defeats the point of taking std::string_view. I think this function could just take a const std::string& instead

source/tools/ActionMap.hpp
```diff
@@ -30,43 +19,55 @@
...
          * For now, it is only required to be non-empty.
          *
-         * @param name attempting to be entered
-         * @return whether name is valid
+         * @param name to validate
+         * @return True if name is valid, false otherwise
          */
         static bool IsValidName(std::string_view name)
         {
             return !name.empty();
```
karensuzue, line 62: I think string_view makes sense here

source/tools/ActionMap.hpp
```diff
@@ -78,47 +79,94 @@ namespace cse498
             return mActions.find(std::string(name));
         }
 
+        template <typename T>
+        static const std::remove_cvref_t<T>* AnyPtr(const std::any& value)
```
karensuzue, line 83: This function needs documentation :) It seems to be some sort of wrapper around std::any_cast (??) but now the value you get out is a pointer, and if you try to cast to the wrong type it returns nullptr (because you're passing the address of value in that return statement).... Anyway, I think all of that should be documented

source/tools/ActionMap.hpp
karensuzue, line 1: This is a well-designed tool! Here are some suggestions:
1. Some functions are missing documentation, and the existing documentation doesn't follow the current company standard. This is especially important because you're using a lot of complex techniques :)
2. Maybe include some sample usage comments in the documentation? The code is a bit hard to decipher at first glance
3. TriggerTyped throws away the information from the InvokeError class by just returning std::nullopt when something fails. I would try to preserve some of that information

source/tools/ActionMap.hpp
```diff
@@ -78,47 +79,94 @@ namespace cse498
...
          * Wrap a typed function (R(Args...)) into a single uniform callable:
          *   invoker(vector<any>) -> any
          *
-         * This is the core trick that lets one map store different signatures safely.
+         * This allows actions with different signatures to be stored in the same map.
          */
         template <typename R, typename... Args>
-        static std::function<std::any(const std::vector<std::any>&)>
+        static std::function<std::expected<std::any, InvokeError>(const std::vector<std::any>&)>
         MakeInvoker(std::function<R(Args...)> func)
```
karensuzue, line 96: Nice! Honestly a lot of this variadic template stuff hurts my head, but I think what you did should work well

karensuzue (approved):

karensuzue:
Just a heads up, I had to make changes to your WorldBase, DemoSimpleWorldG2, and test files because they introduced changes that broke the other groups' demos. Specifically, WorldBase creates a default PlayerAgent, but that class does not actually override SelectAction(), it just asserts false (so when some other group uses MazeWorld, it calls WorldBase's Run function, which calls SelectAction() on all agents).

ZodiacalBee:
> Just a heads up, I had to make changes to your WorldBase, DemoSimpleWorldG2, and test files because they introduced changes that broke the other groups' demos. Specifically, WorldBase creates a default PlayerAgent, but that class does not actually override SelectAction(), it just asserts false (so when some other group uses MazeWorld, it calls WorldBase's Run function, which calls SelectAction() on all agents).

ahhhhh Yeah.. we thought a lot about how to change that because it felt awkward. I feel we need to take the player agent out of the vector of agents owned by the world entirely. We talked a lot about this, and I missed that potential interaction.

karensuzue:
> > Just a heads up, I had to make changes to your WorldBase, DemoSimpleWorldG2, and test files because they introduced changes that broke the other groups' demos. Specifically, WorldBase creates a default PlayerAgent, but that class does not actually override SelectAction(), it just asserts false (so when some other group uses MazeWorld, it calls WorldBase's Run function, which calls SelectAction() on all agents).
> 
> ahhhhh Yeah.. we thought a lot about how to change that because it felt awkward. I feel we need to take the player agent out of the vector of agents owned by the world entirely. We talked a lot about this, and I missed that potential interaction.

For now I'll just merge the temporary fixes I made, so don't worry about it! i'll let you guys sort it out later haha

PR #76, opened 2026-03-26
Group16 initial module and updated classes

MouseCR:
Our two updated classes are
/source/DataAnalytics/ReplayDriver 
/source/tools/Timer

Our module is:
/source/DataAnalytics/HeatMap
/source/DataAnalytics/AnalyticsManager
/source/DataAnalytics/StatsTracker

Our showcase file is
/source/DataAnalytics/group16moduletest.cpp


source/Analyze/AnalyticsManager.hpp
Luke-Antone, line 1: Interoperability & API: This API is pretty locked to just three specific stats. That makes it harder for other groups to use unless they want these exact metrics. A more general way to log named stats would make this easier to plug into other modules.

Code Design: This class still feels partly unfinished because there are a lot of comments about placeholder logs and future features. It would be better to clean that up so the class has a more clear final purpose.

source/DataAnalytics/AnalyticsManager.cpp
```diff
@@ -0,0 +1,60 @@
...
+ **/
+
+#include "AnalyticsManager.hpp"
+
+namespace cse498 {
+
+/*
+Resets all the logs
+*/
+void AnalyticsManager::Reset() {
```
Luke-Antone: Reset() clears everything at once, which is simple, but it may be too broad for a shared analytics class. It could be useful to support resetting one log at a time too.

source/Analyze/HeatMap.hpp
```diff
@@ -0,0 +1,54 @@
...
+  void FillGrid();   // Internal function used to fill out grid when needed
+  std::vector<std::pair<double, double>> ParseActionLog(cse498::ActionLog log); //Private function that parses the action log for its values
+
+public:
+  /** Constructor
+   * ActionLog of all actions to be parsed
+   * Gridsize you want to see pair<int,int>
+   * Worldmap size pair<double,double>
+   */
+  HeatMap(cse498::ActionLog log, std::pair<int, int> gridsize,
```
Luke-Antone, line 37: The HeatMap API is very tied to ActionLog and only really makes sense for move actions. That could make it harder for other groups to use if their position data comes from somewhere else.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+#include <cmath>
+#include <iostream>
+#include <print>
+#include <utility>
+#include <vector>
+// #include "ActionLog.hpp"
+
+// Constexpr this part
+cse498::HeatMap::HeatMap(ActionLog log, std::pair<int, int> ideal_gridsize,
```
Luke-Antone: This constructor silently replaces bad input with default values. That keeps it running, but it also hides mistakes from the caller. It would be better to make invalid input handling clearer.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,126 @@
...
+  m_single_grid_size.second = total_gridsize.second / ideal_gridsize.second;
+
+  m_grid_values.resize(m_ideal_gridsize.first);
+  for (std::vector<int> &x : m_grid_values) {
+    x.resize(m_ideal_gridsize.second);
+  }
+}
+
+std::vector<std::vector<int>>
+cse498::HeatMap::OutPutHeatMap(std::ostream &output) {
```
Luke-Antone: FillGrid() erases entries from m_locations while it runs, so building the heatmap also changes the stored input data. That makes the class harder to reason about and could cause weird behavior if this is called more than once.

source/Analyze/HeatMap.hpp
```diff
@@ -0,0 +1,54 @@
...
+          std::pair<double, double> total_gridsize);
+  /**
+   * Prints the heatmap out in text form
+   * Takes stream to output somewhere else, deafult std::cout
+   * Returns vector of valid values
+   */
+  std::vector<std::vector<int>> OutPutHeatMap(std::ostream &output = std::cout);
```
Luke-Antone, line 47: This method mixes the data part and the output part together. It would be easier for other groups to use if getting the heatmap data and printing the heatmap were separate things.

source/DataAnalytics/ReplayDriver.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <sstream>
+#include <string>
+#include <thread>
+
+#include "../Interfaces/TrashInterface.hpp"
+#include "../Worlds/MazeWorld.hpp"
+#include "ReplayDriver.hpp"
+
+//Helper function to Replay a move
+void cse498::ReplayDriver::ReplayAction(const cse498::Action  &action, cse498::TrashInterface &agent, MazeWorld &world){
```
Luke-Antone: ReplayDriver is very tied to MazeWorld and TrashInterface, so it does not look very reusable outside this one setup. It would be stronger if it depended on more general interfaces instead of specific classes.

demos/DataAnalytics/group16_main.cpp
```diff
@@ -0,0 +1,54 @@
...
+  log.LogAction(1, "move", cse498::WorldPosition{1, 5},
+                cse498::WorldPosition{2, 5});
+  log.LogAction(1, "move", cse498::WorldPosition{2, 5},
+                cse498::WorldPosition{2, 4});
+  log.LogAction(1, "move", cse498::WorldPosition{2, 4},
+                cse498::WorldPosition{2, 3});
+  log.LogAction(1, "move", cse498::WorldPosition{2, 3},
+                cse498::WorldPosition{2, 3});
+
+  cse498::ReplayDriver driver(std::make_shared<cse498::ActionLog>(log));
```
Luke-Antone, line 46: Replaying actions by redirecting std::cin is a pretty fragile design. It works for the demo, but it tightly couples replay logic to console-style input instead of using a cleaner direct action path.

source/DataAnalytics/ReplayDriver.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * int Delay (how long each second of replay should take)
+ */
+  ReplayDriver(std::shared_ptr<const ActionLog> log, int delay = 500)
+      : m_actionLog(std::move(log)), m_delay(delay) {}
+
+  /**
+   * Set a different action Log if wanted
+   * ActionLog log to be added
+   */
+  void setActionLog(std::shared_ptr<const ActionLog> log) {
```
Luke-Antone: Small design thing, but setActionLog uses a different naming style than the rest of the public methods like ReplayByTimeRange and ReplayFullGame. It would be better to keep the naming consistent.

source/Analyze/StatsTracker.hpp
```diff
@@ -0,0 +1,107 @@
...
+ */
+class StatsTracker {
+public:
+  StatsTracker() = default;
+
+  /*@brief Builds a full dashboard snapshot of all tracked data.
+   * @param analytics Analytics manager that owns the underlying logs.
+   * @return Combined snapshot of numeric series summaries and action summaries.
+   */
+  DashboardSnapshot BuildSnapshot(const AnalyticsManager &analytics) const;
```
Luke-Antone, line 75: This is the cleanest API in the module. Returning a DashboardSnapshot with plain summary structs makes it easier for dashboard or GUI code to use without needing direct access to the logs.

source/Analyze/StatsTracker.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return summary;
+}
+
+/*
+ * Combines all numeric series summaries and counters into one snapshot payload.
+ */
+DashboardSnapshot
+StatsTracker::BuildSnapshot(const AnalyticsManager &analytics) const {
```
Luke-Antone, line 44: BuildSnapshot() is easy to follow, but it is still hardcoded to a fixed list of stats. That means every new metric will require changes here too, which could get harder to maintain later.

demos/DataAnalytics/group16_main.cpp
Luke-Antone, line 1: This is a useful demo, but it feels more like a standalone test than a clear integration example for other groups. It would help to show more directly how another module is supposed to plug into this one.

Luke-Antone (commented):
CODE DOES NOT BUILD!!!

Everything else is assuming that you fix that and then make improvements


Interoperability & API:

The module has some solid pieces, especially StatsTracker, which has a clean and usable output format with DashboardSnapshot. That part should be easy for other groups to integrate with. However, some parts of the module are still tightly coupled to specific implementations. ReplayDriver depends directly on MazeWorld and TrashInterface, which makes it hard to reuse in other setups. HeatMap is also tied closely to ActionLog and assumes move-based data, which may not match how other groups structure their data. AnalyticsManager is simple, but it is currently limited to a few hardcoded stats, which makes it less flexible for broader use.

Code Design:

Overall, the code is readable and organized, but some classes still feel like they are in a partial or prototype state. AnalyticsManager still includes placeholder-style comments and could have a more clearly defined final role. HeatMap mixes computation with output and also mutates its internal data while building the grid, which makes its behavior harder to reason about. ReplayDriver works for the demo, but the use of std::cin redirection for replay is a weak design choice and tightly couples it to a specific input system. StatsTracker is the strongest part of the design, though it is still somewhat hardcoded to specific metrics, which could make future expansion harder.

source/Analyze/AnalyticsManager.cpp
AbbyFranz, line 1: This cpp file is mostly single line functions. For readability it might be better to keep all of these one-liners within a single hpp file.

source/Analyze/AnalyticsManager.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/*
+Returns a const reference to the damage dealt log
+*/
+const DataLog &AnalyticsManager::GetDamageDealtLog() const {
+  return mDamageDealtLog;
+}
+
+void AnalyticsManager::LogAction(int entityId, const std::string& actionType,
```
AbbyFranz, line 109: These last 3 functions are missing header comments.

demos/DataAnalytics/group16_main.cpp
AbbyFranz, line 1: There doesnt seem to be a way to actually run your demo, as no executable seems to be created by cmake. The only specs I could find  seems to be the first comment on the pull request, yet there isnt any information on how to run the demo. Please let me know if I am missing something so that I can run the demo.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
...
+    m_ideal_gridsize.second = 10;
+  }
+
+  if (invalid_pair(m_total_gridsize.first, m_total_gridsize.second)) {
+    std::cout << "HeatMap given negative total gridsize values assign new "
+                 "values for valid heatmap";
+    m_total_gridsize.first = 1000;
+    m_total_gridsize.second = 1000;
+  }
+  if ((m_ideal_gridsize.first / m_total_gridsize.first) > .65 &&
```
AbbyFranz: There are a few magic numbers in this file (like .65). I'd recommend making them const variables and also add comments explaining what the numbers represent.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
+#include "HeatMap.hpp"
+
+#include <cmath>
+#include <iostream>
+// #include <print>
+#include <utility>
+#include <vector>
+// #include "ActionLog.hpp"
+
+// Constexpr this part
```
AbbyFranz: This function won't be able to use constexpr here since there are both I/O operations, vectors, and other functions which are also not constexpr.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
...
+cse498::HeatMap::HeatMap(ActionLog log, std::pair<int, int> ideal_gridsize,
+                         std::pair<double, double> total_gridsize) {
+  m_locations = ParseActionLog(log);
+  m_ideal_gridsize = ideal_gridsize;
+  m_total_gridsize = total_gridsize;
+
+  auto invalid_pair = [](auto a, auto b) { return a <= 0 || b <= 0; };
+
+  if (invalid_pair(m_ideal_gridsize.first, m_ideal_gridsize.second)) {
```
AbbyFranz: This could use structred bindings instead of .first and .second to make the code more readable.

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
+#include "HeatMap.hpp"
+
+#include <cmath>
+#include <iostream>
+// #include <print>
+#include <utility>
+#include <vector>
+// #include "ActionLog.hpp"
+
```
rcm7133, line 8: Functions should have Javadoc standard function documetation

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <iostream>
+// #include <print>
+#include <utility>
+#include <vector>
+// #include "ActionLog.hpp"
+
+// Constexpr this part
+cse498::HeatMap::HeatMap(ActionLog log, std::pair<int, int> ideal_gridsize,
+                         std::pair<double, double> total_gridsize) {
+  m_locations = ParseActionLog(log);
```
rcm7133: Implementation: Uses modern C++ constructs

These member variables could be moved into intializer lists for construction like this:

cse498::HeatMap::HeatMap(ActionLog log, std::pair<int, int> ideal_gridsize,
                         std::pair<double, double> total_gridsize)
    : m_locations(ParseActionLog(log)),
      m_ideal_gridsize(ideal_gridsize),
      m_total_gridsize(total_gridsize)
{
 // Rest of the constructor
}

source/DataAnalytics/HeatMap.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * Creates a Heatmap of given entity locations
+ * This heatmap handles edge values (values on borders) by putting them
+ * in the grid down to the right
+ */
+class HeatMap {
+private:
+  std::vector<std::pair<double, double>> m_locations;   // Locations inputted into the heatmap
+  std::vector<std::pair<double, double>> m_invalid_locations;  // Invalid locations from the inputs based on total grid size
+  std::vector<std::vector<int>> m_grid_values;   // Values of how often player was inside grid
+  std::pair<int, int> m_ideal_gridsize;   // Grid size user wants smaller squares / larger squares
```
AbbyFranz: The naming scheme is confusing and a little difficult to understand. For example, based on the comment, it seems like m_ideal_gridsize is actually the size of a cell rather than the size of the full grid.  Whereas the actual full gridsize (m_total_gridsize) would be made up of the size of the cells times the amount of cells there are. Also, some comments do efficiently describe the variable. For example, m_single_grid_size  has the comment "used only for internal calcuations" but this doesnt explain what the variable actually is, and it's confusing as to if it is a cell vs an actual grid section.

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,92 @@
-// #define CATCH_CONFIG_MAIN
```
rcm7133, line 1: SetActionLog is not tested

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,92 @@
...
-//     CHECK(lines[1].rfind("1,move,",   0) == 0);
-// }
-
-// // Replays by agent
-// TEST_CASE("ReplayDriver: ReplayByAgent filters actions", "[ReplayDriver]") {
-//     std::shared_ptr<cse498::ActionLog> log = MakeSampleLog();
-//     cse498::ReplayDriver driver(log);
-
-//     driver.ReplayByAgent(2);
+  REQUIRE_NOTHROW(driver.ReplayFullGame());
```
rcm7133: We are using cassert for error catching and not throws. This checks to make sure the program doesnt crash, not that behavior is correct

source/Analyze/AnalyticsManager.hpp
rcm7133, line 1: Implementation Module Review:

- No warnings during compilation.
- Proper use of namespaces: HeatMap.cpp and ReplayDriver.cpp are using cse498::<Class name>::<Function Name> for implementation. This is fine and works, but may be a little neater if they are just wrapped in cse498 like AnalyticsManager.cpp is
- Adheres to the C++ standard: No detected undefined behavior. Tested and works on windows and linux.
- Appopriate modularity: Good use of helper functions in the module. Gooid use of structs to pack data which is very important for a data focused module. 
- Uses modern C++ constucts: Good use of std::optional in StatsTracker. HeatMap constuctor should use initializer lists to intialize member variables assuming the structure of setting the member variables and error checking them is staying the same. Good use of std::move and intializer lists in ReplayDrive
- Uses standard library algorithms: There is very little need to use std algorithms in this module so far.
- Pointer handling: Good use of smart pointers. No use of raw pointers. 
- Const correctness: Good use of const for the most part. HeatMap could use const on functions like ParseActionLog and OutputHeatMap could be const.

Testing and Error Checking:

- Unit testing: For the most part there is full coverage of all public functions, ReplayDriver SetActionLog is not tested.
- Edge cases are covered well. 
- Assertions: ReplayDriver tests use REQUIRE_NOTHROW and don't use asserts at all. Although for that class it may be diffucult to implement them meaningfully.
- Error handling: Could test for nullptr in constructor of ReplayDriver and invalid time range like ReplayByTimeRange(5.0, 1.0)

source/Analyze/HeatMap.cpp
AbbyFranz, line 1: This file is missing informative comments. I would go into some detail about why/ the purpose of certain code

source/Analyze/ReplayDriver.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+
+#include "../tools/ActionLog.hpp"
+#include "../Interfaces/TrashInterface.hpp"
+#include "../Worlds/MazeWorld.hpp"
+
+namespace cse498 {
+
+class ReplayDriver {
+private:
```
AbbyFranz, line 16: The private variables and functions should also have some kind of comment explaining them like the public functions do

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+  world.UpdateWorld();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(m_delay));
+
+}
+
+// Replays actions that occurred within a specific time range
+void cse498::ReplayDriver::ReplayByTimeRange(double start_time,
+                                             double end_time) {
+  MazeWorld world;
```
AbbyFranz, line 46: For both ReplayByTimeRange and ReplayFullGame a copy of Mazeworld is created. If these functions are later used to replay a randomly generated world instead of a set world, they will need to be adjusted to take in a copy of the world as an argument, as creating a new one will result in a different map, and the replay would be incorrect.

source/Analyze/ReplayDriver.cpp
AbbyFranz, line 1: Naming is inconsitent in this file, variables like fakeInput, originalBuf, actionsInRange are camelCase, but other variables like start_time and end_time are snake_case

source/tools/Timer.cpp
```diff
@@ -5,96 +5,153 @@
...
+  mLaps.push_back(0.0);
+  mRunning = true;
+  mStart = std::chrono::steady_clock::now();
+  mLapStart = std::chrono::steady_clock::now();
+}
+
+/**
+ * Lap the Timer, stopping the previous lap and immediately starting a new one.
+ */
+void cse498::Timer::lap() {
```
AbbyFranz, line 81: This function has the potential to be misused. There is no check for mRunning in this function so a user could call lap on the timer when it is not running, and incorrect values could be stored.

source/tools/Timer.cpp
```diff
@@ -5,96 +5,153 @@
...
+  oss << mName << " [" << (mRunning ? "RUNNING" : "STOPPED")
+      << "]: " << std::setfill('0') << std::setw(2)
+      << static_cast<int>(time / 60) << ":" << std::setw(6) << std::fixed
+      << std::setprecision(3) << std::fmod(time, 60.0);
+
+  if (withLaps) {
+    for (auto [i, lapTime] : std::ranges::views::enumerate(mLaps)) {
+      oss << "\n  Lap " << (i + 1) << ": " << std::setfill('0') << std::setw(2)
+          << static_cast<int>(lapTime / 60) << ":" << std::setw(6) << std::fixed
+          << std::setprecision(3) << std::fmod(lapTime, 60.0);
```
AbbyFranz: 60.0 is a magic number and could be stored within its own const variable

AbbyFranz (commented):
Summary

Code Design:
-Informative identifiers: Good for the most part, there are only a few variables in HeatMap that are potentially confusing.
-Avoids Magic Numbers: A few classes have some magic numbers.
-Intuitive and functional interface: Looks good.
-Informative Comments: HeatMap, ReplayDriver, and StatsTracker are all missing informative comments (Specifically inline informative comments, the header doxygen style comments are all fine).
-Consistent Style: Style across files is inconsistent for both variable and function names. For example files in the DataAnalytics folders use snake_case for variables and member variables, and PascalCase for functions. However, the timer class uses camelCase for variables and functions and PascalCase for member variables. Also function documentation differs slightly within files. Some files keep their doxygen style header comments in the hpp file, while others have them in the cpp file. It would be best to decide to keep them in one or the other so users can always know where to find the documentation.
-Readability: Mostly good, the suggestion about keeping header comments consistent will help with readability as well. 

Advanced C++:
This is good - The classes incorporate advanced topics in areas where they are able to.
StatsTracker makes good use of optional
HeatMap has correct usage of lambdas
Timer uses ranges correctly

source/DataAnalytics/AnalyticsManager.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  /*
+  Adds a new damage dealt value to the damage dealt log
+  */
+  void LogDamageDealt(double damage);
+
+  /*
+  Returns a const reference to the health log
+  */
+  const DataLog &GetHealthLog() const;
```
gudimaya: Advanced C++ Review:

Consider adding [[no discard]] to these getters. Prevents ignoring return values

[[nodiscard]] const DataLog& GetHealthLog() const noexcept;
[[nodiscard]] const DataLog& GetEnemiesTrackedLog() const noexcept;
[[nodiscard]] const DataLog& GetDamageDealtLog() const noexcept;
[[nodiscard]] const ActionLog& GetActionLog() const noexcept;

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
...
+  m_ideal_gridsize = ideal_gridsize;
+  m_total_gridsize = total_gridsize;
+
+  auto invalid_pair = [](auto a, auto b) { return a <= 0 || b <= 0; };
+
+  if (invalid_pair(m_ideal_gridsize.first, m_ideal_gridsize.second)) {
+    std::cout << "HeatMap ideal gridsize negative assign new values for valid "
+                 "heatmap";
+    m_ideal_gridsize.first = 10;
```
gudimaya: Advanced C++ Review:

Consider adding constexpr or inline constexpr for defaults since there are many constants in the code.

inline constexpr int DefaultIdealGridWidth = 10;
inline constexpr int DefaultIdealGridHeight = 10;
inline constexpr double DefaultTotalGridWidth = 1000.0;
inline constexpr double DefaultTotalGridHeight = 1000.0;
inline constexpr double GridWarningThreshold = 0.65;

source/DataAnalytics/AnalyticsManager.cpp
```diff
@@ -0,0 +1,75 @@
...
+  mHealthLog.Clear();
+  mEnemiesTrackedLog.Clear();
+  mDamageDealtLog.Clear();
+  mActionLog.Clear();
+}
+
+/*
+Adds a new health value to the health log
+*/
+void AnalyticsManager::LogHealth(double health) { mHealthLog.Add(health); }
```
gudimaya: Implementation Quality Review:

Implementation quality is good overall since the methods are simple, consistent, and easy to maintain. The main improvement would be adding stronger validation or safer typing, especially for string-based action types.

for void AnalyticsManager::LogHealth(double health), you could consider adding a safety measure to ensure health isn't a negative value.

demos/DataAnalytics/group16_main.cpp
```diff
@@ -0,0 +1,54 @@
...
+  analytics.LogHealth(80.0);
+  analytics.LogEnemiesTracked(3);
+  analytics.LogDamageDealt(25.0);
+
+  analytics.LogHealth(100.0);
+  analytics.LogEnemiesTracked(1);
+  analytics.LogDamageDealt(50.0);
+
+  cse498::ActionLog log;
+  log.LogAction(1, "move", cse498::WorldPosition{0, 0},
```
gudimaya, line 25: Implementation Quality Review:

Instead of using string values for actions like "move", "attack", "heal", utilize an enum class called ActionType. Just a suggestion, may not be necessary

tests/DataAnalytics/AnalyticsManagerTest.cpp
alexelkins23, line 1: I am not seeing any negative testing in this file, I would like to see some test cases with invalid updates to ensure the class can handle invalid data gracefully.

tests/tools/HeatMapTests.cpp
alexelkins23, line 1: I like the test cases more in this file, there is good use of invalid data with negative values, but I would also like to see very large values that may be too big for the grid.

tests/tools/ReplayDriverTests.cpp
alexelkins23, line 1: This file also suffers from a lack of diverse test cases, I would like to see some invalid data and a more robust verification that the output is actually what we want, as mentioned by other reviewers

tests/tools/StatsTrackerTest.cpp
alexelkins23, line 1: Again I would like more negative testing, but I do like the robust verification for valid data.

tests/tools/TimerTest.cpp
alexelkins23, line 1: I like the testing in this file, but I would like a little more attention given to the lap function and its interaction with mRunning as other reviewers mentioned.

demos/DataAnalytics/group16_main.cpp
alexelkins23, line 1: This demo does show how the class could be used, but not in the context of the other modules. I would suggest creating a mock world that could inherit from BaseWorld or MazeWorld and use the AnalyticsManager to track actions and then print them after the game is stopped.

source/DataAnalytics/HeatMap.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  if (invalid_pair(m_total_gridsize.first, m_total_gridsize.second)) {
+    std::cout << "HeatMap given negative total gridsize values assign new "
+                 "values for valid heatmap";
+    m_total_gridsize.first = 1000;
+    m_total_gridsize.second = 1000;
+  }
+  if ((m_ideal_gridsize.first / m_total_gridsize.first) > .65 &&
+      m_ideal_gridsize.second / m_total_gridsize.second > .65) {
+    std::cout << "Warning Heatmap ideal gridsize is to close or larger in size "
```
alexelkins23: I like this warning for data that isn't invalid, but may be unhelpful. It would be good to have more things like this and, like other reviewers mentioned, more robust error handling for truly invalid data like asserts.

alexelkins23 (commented):
From a interoperability standpoint I also think there are a couple changes we could make for better integration. I think the specific analytics that we collect should be defined by the world, and the world would tell this module what it wants tracked, then they could be stored as a key-value map. This would make the module a little more flexible, so we aren't tracking unused or unwanted metrics. And like other reviewers suggested, making the actions a enum, which would be shared across the worlds to reduce ambiguity.

For testing and error checking, there is a good foundation, but there are more edge cases and negative cases that need to be addressed.

source/tools/Timer.cpp
ncharters: Thank you for catching this! I will be fixing that right now

source/tools/Timer.cpp
ncharters: Fixed, thank you

tests/tools/TimerTest.cpp
ncharters: Agreed, I've added some more tests to make sure that attempting to lap while stopped does not lap the timer.

source/Analyze/ReplayDriver.cpp
MouseCR: Yeah this class is going to be updated more in the future. For now it just kind of works as a example. Once we get more information on how the game functions well update it again.

source/DataAnalytics/HeatMap.cpp
MouseCR: Yeah I don't remember why we put that there I think that was just a accident.

source/DataAnalytics/HeatMap.cpp
MouseCR: I was using m_locations as unprocessed input so erasing each location after inserting prevents the location from being counted multiple times if another actionLog is added and more locations are added to the Heatmap. I agree that I did not explain this in a comment.

source/Analyze/HeatMap.hpp
MouseCR: Our idea for our module is that action log will track all actions taken by all agents and therefore heatmap will always be available to use because each agent will have a actionlog.

source/DataAnalytics/HeatMap.hpp
MouseCR: Yeah I agree they make sense in my head but are hard to understand for people who didn't make this. They're updated and should be better now.

source/Analyze/HeatMap.hpp
MouseCR: Made it public.

source/Analyze/ReplayDriver.cpp
MouseCR: Thanks for catching that its fixed now

source/Analyze/ReplayDriver.hpp
MouseCR: Thanks for catching that, its fixed now

source/DataAnalytics/ReplayDriver.hpp
MouseCR: Thanks for catching that, its fixed now

source/DataAnalytics/HeatMap.cpp
MouseCR: Yeah, updated the error handling to be better.

source/DataAnalytics/HeatMap.cpp
MouseCR: Ended up getting rid of the magic numbers entirely but this was a good idea

source/DataAnalytics/HeatMap.cpp
MouseCR: Ended up getting rid of the magic numbers entirely but this was a good idea

source/DataAnalytics/HeatMap.cpp
MouseCR: Good change, it's updated now

source/DataAnalytics/HeatMap.cpp
MouseCR: Changed to have better error handling

source/Analyze/StatsTracker.cpp
murra381: Thank you for your feedback. I will be changing this in the future to account for these growing constraints with my class. Those were a specific group of stats we decided would be good for a summary.

source/Analyze/StatsTracker.hpp
murra381: Thank you!

demos/DataAnalytics/group16_main.cpp
MouseCR: Yeah for now this class is build of the simple game given. The only way I could find to make this work was forcing inputs. This class will be updated in the future to work on the full game in a better way.

source/Analyze/HeatMap.cpp
MouseCR: Added comments in thanks for catching that.

source/Analyze/HeatMap.cpp
MouseCR: Updated documentation throughout

source/DataAnalytics/HeatMap.cpp
MouseCR: Good point, I have added them in.

tests/tools/HeatMapTests.cpp
MouseCR: Yeah I didn't see much value of having a value significantly larger but I have added another test case in with large values.

source/Analyze/ReplayDriver.hpp
```diff
@@ -0,0 +1,55 @@
...
+  std::shared_ptr<const ActionLog> m_actionLog;   //action log of all actions to be replayed
+  int m_delay;   //delay for how long each move in the replay should take (in milliseconds)
+  void ReplayAction(const cse498::Action &action, cse498::TrashInterface &agent, MazeWorld &world);   //helper function to replay a single action
```
karensuzue, line 19: Your function currently takes in a TrashInterface agent and a MazeWorld directly, which limits modularity. Ideally, you want to design it so you can plug in any type of interface/agent and any type of world

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
-    Actions.push_back(action);
+    action.Timestamp      = mCurrentTime;
+    action.Position       = position;
+    action.NewPosition    = newPosition;
+    action.SequenceNumber = mNextSequenceNumber++;
+    mActions.push_back(action);
 }
 
 void ActionLog::UpdateTime(double newTime) {
     assert(newTime >= 0.0 && "Time must be non-negative");
```
karensuzue, line 36: As mentioned in https://github.com/CSE498/Spring2026-CompanyA/pull/36#discussion_r2896004926, we strongly discourage using asserts for input validation since they can be disabled at compile time.

karensuzue: There are a few easy ways you could handle this. You could return a boolean to indicate whether the operation succeeded, or you could simply clamp the time value. Another option is to use std::expected, which would also give you a chance to experiment with more modern C++ :)

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
+    mCurrentTime = newTime;
 }
 
 const std::vector<Action>& ActionLog::GetActions() const {
-    return Actions;
+    return mActions;
 }
 
 std::vector<Action> ActionLog::GetActionRange(double startTime, double endTime) const {
     assert(startTime <= endTime && "startTime must be <= endTime");
```
karensuzue, line 45: Unaddressed reviewer comment: https://github.com/CSE498/Spring2026-CompanyA/pull/36#discussion_r2896004926

Please make sure to respond to all reviewer comments this time, even if you disagree with a proposed change (feel free to state why!) It's simply good etiquette to do so :) I know this isn't one of the classes you selected for submission, but since it's still used by other parts of the project I'd recommend improving it whenever you can

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
     std::vector<Action> result;
-    for (const auto& action : Actions) {
-        if (action.Timestamp >= startTime && action.Timestamp <= endTime) {
-            result.push_back(action);
-        }
-    }
+    auto inRange = [startTime, endTime](const Action& a) {
+        return a.Timestamp >= startTime && a.Timestamp <= endTime;
+    };
+    std::copy_if(mActions.begin(), mActions.end(), std::back_inserter(result), inRange);
```
karensuzue, line 51: Nice use of copy_if and lambdas! startTime and endTime are passed by value here, but since they're doubles that should be fine

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
+    return static_cast<int>(mActions.size());
+}
+
+std::optional<int> ActionLog::GetMostActiveEntity() const {
+    if (mActions.empty()) {
+        return std::nullopt;
+    }
+
+    std::unordered_map<int, int> counts;
+    std::for_each(mActions.begin(), mActions.end(),
```
karensuzue, line 71: I'd just use a regular range‑based for loop here :) there isn't much reason to use for_each

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
+        [&counts](const Action& a) { ++counts[a.EntityId]; });
```
karensuzue, line 72: I think it would be more efficient to make the map a member variable, so you don't have to reconstruct it every time the function is called

source/tools/ActionLog.cpp
```diff
@@ -11,67 +11,84 @@
...
-
-    Actions.push_back(action);
+    action.Timestamp      = mCurrentTime;
+    action.Position       = position;
+    action.NewPosition    = newPosition;
+    action.SequenceNumber = mNextSequenceNumber++;
+    mActions.push_back(action);
 }
 
 void ActionLog::UpdateTime(double newTime) {
```
karensuzue, line 35: Just a heads up, the way this function is implemented doesn't enforce monotonic time :) You can update the time in any order. Make sure that's intentional

source/Analyze/ReplayDriver.hpp
```diff
@@ -0,0 +1,55 @@
...
+public:
+ /**
+   * Constructs a ReplayDriver with an action log and optional replay delay.
+   *
+   * @param log Shared pointer to the ActionLog containing actions to replay.
+   * @param delay Delay between replayed actions in milliseconds.
+   *              Defaults to 500.
+   */
+  ReplayDriver(std::shared_ptr<const ActionLog> log, int delay = 500)
+      : m_actionLog(std::move(log)), m_delay(delay) {}
```
karensuzue, line 29: Some interesting choices here - you're passing a shared_ptr by value, which makes a copy of the pointer, then calling the shared_ptr move constructor. That's probably okay, but I think it's cleaner to just do:
```
ReplayDriver(const std::shared_ptr<const ActionLog>& log, int delay = 500)
    : m_actionLog(log), m_delay(delay) {}
```
I don't think this is necessarily more efficient, but it is a bit easier to read. It simply copies `log`, and there's no need to move it.

Also, another thing is why shared_ptr? Do you expect multiple objects to share ownership of the same log? Should the driver own the log, or should the log outlive it? I'm not totally sure what ownership model you're aiming for, but it's worth making sure the choice has a clear purpose :)

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "../Interfaces/TrashInterface.hpp"
+#include "../Worlds/MazeWorld.hpp"
+#include "ReplayDriver.hpp"
+
+// Helper function to Replay a move
+void cse498::ReplayDriver::ReplayAction(const cse498::Action &action,
+                                        cse498::TrashInterface &agent,
+                                        MazeWorld &world) {
+  char move = 'x';
+
```
karensuzue, line 16: I think some documentation for lines 15-26 would help here :)

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (action.Position.X() > action.NewPosition.X())
+    move = 'a';
+  else if (action.Position.X() < action.NewPosition.X())
+    move = 'd';
+  else if (action.Position.Y() > action.NewPosition.Y())
+    move = 'w';
+  else if (action.Position.Y() < action.NewPosition.Y())
+    move = 's';
+  else
+    return;
```
karensuzue, line 26: I wonder if it would be better to return a bool or use std::expected to indicate success or failure instead of returning void.

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  size_t action_id = agent.SelectAction(world.GetGrid());
+  int result = world.DoAction(agent, action_id);
+  agent.SetActionResult(result);
+
+  std::cin.rdbuf(original_buf);
+  world.UpdateWorld();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(m_delay));
+
```
karensuzue, line 40: Are you planning to look at ActionType as well? It seems like this function only replays movement actions

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (action.Position.X() > action.NewPosition.X())
+    move = 'a';
+  else if (action.Position.X() < action.NewPosition.X())
+    move = 'd';
+  else if (action.Position.Y() > action.NewPosition.Y())
+    move = 'w';
+  else if (action.Position.Y() < action.NewPosition.Y())
+    move = 's';
+  else
```
karensuzue, line 25: It looks like the function only considers movement in a single direction (and no diagonal movement), which is fine given how our system is set up...

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return;
+
+  std::istringstream fake_input(std::string(1, move) + "\n");
+  std::streambuf *original_buf = std::cin.rdbuf();
+  std::cin.rdbuf(fake_input.rdbuf());
```
karensuzue, line 30: I’m guessing you’re trying to simulate keyboard input here?  

It actually took me a while to understand what you were trying to do… so yeah, some documentation is very much needed :)

source/Analyze/ReplayDriver.cpp
```diff
@@ -0,0 +1,65 @@
...
+  world.UpdateWorld();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(m_delay));
+
+}
+
+// Replays actions that occurred within a specific time range
+void cse498::ReplayDriver::ReplayByTimeRange(double start_time,
+                                             double end_time) {
+  MazeWorld world;
```
karensuzue, line 46: You shouldn't be hard coding MazeWorld here! I know you mentioned it's just an example, but that's something the demo should handle. This module is meant to be a general purpose tool. I know that's not an easy task, but taking the time to write modular/reusable code is an important skill to develop

source/tools/Timer.cpp
```diff
@@ -5,96 +5,160 @@
...
 
 /**
- * Constructor for a Timer.
+ * Timer constructor.
+ * @param name The name of the Timer.
+ * @param startRunning Whether the Timer should start running. Defaults to true.
  */
-cse498::Timer::Timer()
-{
+cse498::Timer::Timer(std::string name, bool startRunning) {
```
karensuzue, line 21: if you want startRunning to default to true, you need `bool startRunning = true`

source/tools/Timer.cpp
```diff
@@ -5,96 +5,160 @@
...
+  mName = name;
+  mRunning = startRunning;
+  mStart = std::chrono::steady_clock::now();
+  mLapStart = std::chrono::steady_clock::now();
```
karensuzue, line 25: Is it really the best choice to initialize both time points when startRunning is false? That feels a bit misleading, I think! You also end up redoing the same initialization in start(), which isn't a huge issue, but it does make the design feel a little less clean. 
This might be a nice place to consider std::optional, so an unset time point can ACTUALLY be represented as unset :)

source/tools/Timer.cpp
```diff
@@ -5,96 +5,160 @@
...
-        auto startTime = mTimers[timerName].startTime;
-        auto endTime = std::chrono::steady_clock::now();
-        mTimers[timerName].accumulatedDuration += endTime - startTime;
-        mTimers[timerName].isRunning = false;
-    }
+void cse498::Timer::stop() {
+  if (mRunning) {
+    mRunning = false;
+    mElapsed += (std::chrono::steady_clock::now() - mStart);
+    mLaps.at(mLaps.size() - 1) +=
```
karensuzue, line 48: I was really confused for a while about why mLaps is a vector of size 1. You should probably add some documentation explaining that it expands in the .hpp file

source/tools/Timer.cpp
```diff
@@ -5,96 +5,160 @@
...
-        // and mark it as stopped.
-        auto startTime = mTimers[timerName].startTime;
-        auto endTime = std::chrono::steady_clock::now();
-        mTimers[timerName].accumulatedDuration += endTime - startTime;
-        mTimers[timerName].isRunning = false;
-    }
+void cse498::Timer::stop() {
+  if (mRunning) {
+    mRunning = false;
+    mElapsed += (std::chrono::steady_clock::now() - mStart);
```
karensuzue, line 47: std::chrono::steady_clock::now() is called twice in this function

source/tools/Timer.cpp
```diff
@@ -5,96 +5,160 @@
...
 
-bool cse498::Timer::isRunning(const std::string &timerName) const
-{
-    return mTimers.find(timerName) != mTimers.end() ? mTimers.at(timerName).isRunning : false;
+/**
+ * Lap the Timer, stopping the previous lap and immediately starting a new one.
+ */
+void cse498::Timer::lap() {
+  if (mRunning) {
+    mLaps.at(mLaps.size() - 1) +=
```
karensuzue, line 83: You should probably add a check to make sure mLaps is the correct size and initialized with the right value (0.0) before you start adding to it

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,175 @@
...
 
-    return log;
+  return log;
 }
 
-// Tests for ReplayDriver
-TEST_CASE("ReplayDriver: IsActionLogSet()", "[ReplayDriver]") {
-    cse498::ReplayDriver driver;
-    CHECK(driver.IsActionLogSet() == false);
+std::string CaptureStdout(const std::function<void()>& fn) {
```
karensuzue, line 61: I would recommend adding some documentation here, but otherwise this is a good idea!

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,175 @@
...
+          "[ReplayDriver]") {
+  auto log = MakeFullGameLog();
+  cse498::ReplayDriver driver(log, 0);
 
-//     auto lines = ReadAllLines(file);
-//     CHECK(lines.empty());
-// }
+  REQUIRE_NOTHROW(CaptureStdout([&]() {
+    driver.ReplayFullGame();
+  }));
```
karensuzue, line 110: I'm pretty sure ReplayFullGame() does not throw exceptions, especially now that our company has decided to stop using them. Because of that, this test will always pass, so it isn't very meaningful.

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,175 @@
...
-// TEST_CASE("ReplayDriver: ReplayByTimeRange filters actions by timestamp", "[ReplayDriver]") {
-//     std::shared_ptr<cse498::ActionLog> log = MakeSampleLog();
-//     cse498::ReplayDriver driver(log);
+TEST_CASE("ReplayDriver ReplayByTimeRange replays only selected time window",
+          "[ReplayDriver]") {
+  auto log = MakeTimedLog();
+  cse498::ReplayDriver driver(log, 0);
 
-//     driver.ReplayByTimeRange(1.0, 2.0);
+  REQUIRE_NOTHROW(CaptureStdout([&]() {
```
karensuzue, line 118: Misuse of REQUIRE_NOTHROW

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,175 @@
...
+          "[ReplayDriver]") {
+  auto log = MakeTimedLog();
+  cse498::ReplayDriver driver(log, 0);
 
-//     auto lines = ReadAllLines(file);
-//     REQUIRE(lines.size() == 2);
-//     CHECK(lines[0].rfind("2,attack,", 0) == 0);
-//     CHECK(lines[1].rfind("1,move,",   0) == 0);
-// }
+  REQUIRE_NOTHROW(CaptureStdout([&]() {
```
karensuzue, line 128: Another misuse of REQUIRE_NOTHROW. Try to validate the actual functionality instead. You have an action log, and you already have a helper that captures stdout into a string, so use those to check observable behavior.

This is also why I often encourage returning bool or using std::expected where appropriate, it gives you something meaningful to check for in tests. 

Also right now, these tests mostly just verify that nothing crashes. That's not always the same as "my function works as intended" :)

karensuzue: Though I have to say... the way these tests and your code are set up makes it pretty confusing to check observable behavior. But I do think it's doable. As I understand it:
1. ReplayFullGame() creates a TrashInterface and then calls ReplayAction() for each action in the log.
2. ReplayAction() infers a move key, stuffs it into a fake input stream, calls agent.SelectAction()
3. Inside SelectAction(), it calls DrawGrid(), which prints the board before reading the input. That output should then be captured by the helper function...

Okay, this is a bit problematic because the observable behavior is HIGHLY dependent on the implementation of TrashInterface::SelectAction. Which means things wouldn't work the same if a different interface was used.

tests/tools/ReplayDriverTests.cpp
```diff
@@ -1,168 +1,175 @@
...
-// TEST_CASE("ReplayDriver: ReplayByActionType filters actions", "[ReplayDriver]") {
-//     std::shared_ptr<cse498::ActionLog> log = MakeSampleLog();
-//     cse498::ReplayDriver driver(log);
+  cse498::ReplayDriver driver(empty_log, 0);
 
-//     driver.ReplayByActionType("move");
+  std::string empty_output = CaptureStdout([&]() {
+    driver.ReplayFullGame();
+  });
+  REQUIRE(empty_output.empty());
```
karensuzue, line 155: Okay, I see that you're checking observable behavior here, which is good! But this test is still heavily dependent on the fact that your code uses TrashInterface specifically

karensuzue:
Demo built successfully with 
`g++ -std=c++23 -I./source -I./source/DataAnalytics -I./source/tools source/DataAnalytics/Group16_main.cpp source/DataAnalytics/AnalyticsManager.cpp source/DataAnalytics/HeatMap.cpp source/DataAnalytics/ReplayDriver.cpp source/tools/DataLog.cpp source/tools/ActionLog.cpp source/tools/Timer.cpp -o group16moduletest`

Seeing the demo run is really cool! Nice work figuring out the replays via forcing inputs. I'm not sure how well we'll be able to integrate the GUI/World groups with this, but having it working in the console is still a solid step forward. 

For the next stage, please feel free to reach out to me or any of the instructors directly (DM) if you need help with the integration. I think there's quite a bit of refactoring to be done

source/Analyze/AnalyticsManager.hpp
karensuzue, line 1: I agree with Luke here: to add more stats (beyond what's currently available - health, enemies, damage, action...etc.), you'll need to dig into this code, which isn't ideal. There are ways to make this more general so you can simply "plug in" a statistic. Then you could have a more general Reset function that takes a stat's name and resets its corresponding log, for example.

I highly encourage you guys to look into this.

source/Analyze/AnalyticsManager.cpp
```diff
@@ -0,0 +1,128 @@
...
+ * ActionLogs for the project. This class will be called by the game to log data
+ * and actions which will be used by the StatsTracker class.
+ **/
+
+#include "AnalyticsManager.hpp"
+
+namespace cse498 {
+
+/*
+Resets the health log
```
karensuzue, line 15: The documentation style doesn't seem to align with the company standard, and the style is inconsistent across the group.

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <cmath>
+#include <iostream>
+#include <stdexcept>
+#include <utility>
+#include <vector>
+
+cse498::HeatMap::HeatMap(ActionLog log, std::pair<int, int> ideal_heatmap_size,
+                         std::pair<double, double> map_size)
+    : m_locations(ParseActionLog(log)),
```
karensuzue, line 11: Member variable names don't follow the company standard

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+  const auto [heatmap_rows, heatmap_cols] = m_ideal_heatmap_size;
+  const auto [map_width, map_height] = m_map_size;
+  auto& [cell_width, cell_height] = m_single_grid_size;
+
+  auto invalid_pair = [](auto a, auto b) { return a <= 0 || b <= 0; };
+
+  // Both the heatmap dimensions and map dimensions must be positive
+  // before any cell sizing can be computed
+  if (invalid_pair(heatmap_rows, heatmap_cols)) {
+    throw std::invalid_argument(
```
karensuzue, line 24: The company has agreed not to use exceptions

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  if (invalid_pair(map_width, map_height)) {
+    throw std::invalid_argument(
+        "HeatMap: total_gridsize must have positive dimensions");
+  }
+
+  // Each heatmap cell represents this much space in the world map
+  cell_width = map_width / heatmap_rows;
+  cell_height = map_height / heatmap_cols;
```
karensuzue, line 35: I feel like... this should be the opposite?
cell width = map_width / heatmap_cols
cell height = map_height / heatmap_rows

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+  // Print any coordinates that fell outside the configured map bounds
+  for (const auto& [x, y] : m_invalid_locations) {
+    output << x << "," << y << std::endl;
+  }
+
+  return m_grid_values;
+}
+
+std::vector<std::vector<int>> cse498::HeatMap::FillGrid() {
+  const auto [map_width, map_height] = m_map_size;
```
karensuzue, line 80: This is a good use of structured bindings :)

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+  const auto [cell_width, cell_height] = m_single_grid_size;
+
+  // A location is valid if it is inside the map bounds and maps to a
+  // non-negative grid index
+  auto in_bounds = [map_width, map_height](double x, double y, double grid_x,
+                                           double grid_y) {
+    return x < map_width && y < map_height &&
+           grid_x >= 0 && grid_y >= 0;
+  };
```
karensuzue, line 89: You might want to check for x < 0, y < 0, and the upper bound for grid_x and grid_y

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+  // locations that were already added to the heatmap
+  for (auto it = m_locations.begin(); it != m_locations.end();) {
+    const auto& [x, y] = *it;
+
+    // Convert world coordinates into heatmap grid indices
+    double temp_grid_x = std::floor(x / cell_width);
+    double temp_grid_y = std::floor(y / cell_height);
+
+    if (in_bounds(x, y, temp_grid_x, temp_grid_y)) {
+      m_grid_values[temp_grid_x][temp_grid_y] += 1;
```
karensuzue, line 101: In many 2D vector grids, the convention is grid[row][col], and usually row = y, col = x. Make sure you're indexing into the right locations!

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+    } else {
+      // Keep track of out-of-bounds inputs separately for later inspection
+      m_invalid_locations.push_back(*it);
```
karensuzue, line 104: This is a good idea! :) Though maybe it would be better to let the user of the class decide whether they want to store the invalid inputs or not, since they could easily pile up

source/Analyze/HeatMap.cpp
```diff
@@ -0,0 +1,137 @@
...
+           grid_x >= 0 && grid_y >= 0;
+  };
+
+  // Consume each pending location once so later calls do not double-count
+  // locations that were already added to the heatmap
+  for (auto it = m_locations.begin(); it != m_locations.end();) {
+    const auto& [x, y] = *it;
+
+    // Convert world coordinates into heatmap grid indices
+    double temp_grid_x = std::floor(x / cell_width);
```
karensuzue, line 97: I think this should be an int?

karensuzue (approved):

karensuzue:
Hi! I just noticed a potential issue with your Timer test cases. For a test case like this:

```
TEST_CASE("Timer Constructor", "[Timer]") {
  // Test that single parameter constructor defaults the Timer to running.
  cse498::Timer Timer("Test1");
  REQUIRE(Timer.isRunning());
  REQUIRE(Timer.elapsed() > 0);
}
```

They were failing on some people's machines. I suspect this is because the test can run so quickly that the elapsed time is still effectively 0 by the time that last assertion is checked. Could you find a different way to test this? Perhaps with `std::this_thread::sleep_for(std::chrono::seconds(5))`?

PR #77, opened 2026-03-31
Item Class For Merge Ease

ZodiacalBee:
This is a class that is needed by both Group2 and Group15 and we have the same implementation. This will help the merge. There isn't a whole lot to comment on outside of implementation design since it is pretty simple. Most of this pull is the test file which just ensures some very basic functionality


tests/tools/GenerativeWorld/ItemTest.cpp
```diff
@@ -0,0 +1,90 @@
+/**
+ * @file ItemTest.cpp
+ * @author Group 2
+ */
+
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "core/item/Item.hpp"
```
nahumj: Shouldn't this be:
```
#include "../../../source/core/item/item.hpp"
```
And shouldn't the name of the file be Item.hpp (with an uppercase I)?

nahumj (changes requested):

tests/tools/GenerativeWorld/ItemTest.cpp
nahumj: image

ZodiacalBee: wait how did that change? It was Item.hpp? so weird... I explicitly changed this..

nahumj (approved):

PR #78, opened 2026-04-15
Fixed Timer Tests to not fail on certain builds

ncharters:
Test cases which included 'REQUIRE(Timer.elapsed() > 0)' were failing on some builds due to taking really close to 0 seconds. This updates the tests so that they use the advanceTime() function in order to be significantly over 0 only in the case that the timer is running.


karensuzue (approved):
Thank you!!

PR #82, opened 2026-04-18
Group16 - AgentBase changes

Ebizi:
Hey! Making a small pull request with my AgentBase changes in order to implement an ActionLog within each agent, was recommended to start up a small pull request with this change in place once it was pushed


karensuzue (approved):
This looks good!

PR #83, opened 2026-04-19
Implemented Interaction for agents

ZodiacalBee:
This is for GUIs to tie in and worlds to tie in for calling functions when "E" is pressed or when "left click" is pressed. 
I may forget to change to merge ready. I'm just waiting on the response from discord about left click / E

NOTE: Just read the FIRST COMMIT. The second clang-format commit changed a lot more than I thought within the 4 files that I modified? Was I not supposed to do that?


karensuzue:
I did a git diff and it looks like your commit only touched those 4 files, so no worries!

karensuzue (approved):
This looks good!

PR #86, opened 2026-04-20
Stats logging & menu page

ncharters:
Adds the logging of two statistics - Damage dealt to enemies & Enemies killed - as well as a menu page to display those statistics.


karensuzue (approved):

source/Agents/Classic/Enemy.cpp
alexelkins23, line 1: I like that you integrated the damage and death analytics in the enemy agent class since it has all the necessary information and can pass it to the world when needed.

source/Analyze/AnalyticsManager.cpp
alexelkins23, line 1: This does a good job of providing a simple API with the necessary functions for the world/GUI to use. It does not expose unnecessary complexity, just the data that is relevant in an organized fashion.

source/core/WorldBase.hpp
alexelkins23, line 1: I like that the AnalyticsManager was added to WorldBase so both worlds immediately get access and since the agents collect the data, they should be able to use it immediately without much refactoring.

alexelkins23 (commented):
Code Design: I really like the design of this module, it is well integrated with the other modules and they should be able to start using it without difficulty. The code has a consistent style and does not use magic numbers/values.

Advanced C++: As with the main PR, I don't see many uses of advanced C++ features, I don't see much use for value semantics, constexpr, ranges, or lambdas. I made a possible suggestion with templates on the main PR.

source/Analyze/AnalyticsManager.hpp
```diff
@@ -14,33 +14,49 @@
...
 
-    /*
-    Resets the health log
-    */
-    void ResetHealthLog();
+    /**
+     * Resets the current run stats to their default values.
+     */
+    void ResetCurrentRunStats() { mCurrentRunStats = {}; }
+
```
paulbui081, line 46: Make sure to document what LogRunDamage(...) and LogRunEnemiesKilled(...) do

tests/DataAnalytics/AnalyticsManagerTest.cpp
paulbui081, line 1: ### Testing and Error checking
Compiling and running the test cases for this file returns no error cases. Looking through the changes and the overall structure of the file contents, these test make sure to properly test through a majority of the public member functions. 

It does need to test the following functions though:
- LogRunDamage(...)
- LogRunEnemiesKilled(...)

Other than that minor nitpick, everything is looking good. Well Done!

demos/DataAnalytics/Group16_main.cpp
paulbui081, line 1: ### Interoperability and API
Not much has changed compared to the last pull request, but from what I'm seeing this demo module does a good job at showcasing how the DataAnalytics class could be used to gather information of player behavior and movement. However, it would of been nice to see how this class could be integrated with the other group's module as this seems like it's only showing how this module could be used by itself, not with the other modules. The API interface is well documented and the names of the functions are descriptive enough to understand its general purpose.

source/Agents/Classic/Enemy.hpp
```diff
@@ -55,6 +55,16 @@ class Enemy : public AgentBase {
      */
     void ResetGoldDropClaim() { mGoldClaimed = false; }
 
+    /**
+     * Override of TakeDamage to log damage dealt to this enemy in the analytics manager.
+     */
+    void TakeDamage(double amount) override;
```
paulbui081, line 61: Be sure to add a comment documenting what 'amount' represents.

source/Analyze/AnalyticsManager.cpp
```diff
@@ -35,32 +40,21 @@ void AnalyticsManager::ResetActionLog() { mActionLog.Clear(); }
...
-    }
-    mHealthLog.Add(health);
-}
 
 /*
-Adds a new enemy count value to the enemies tracked log
+Adds a new enemy count value to the enemies killed log
 */
-void AnalyticsManager::LogEnemiesTracked(double count_enemies) {
+void AnalyticsManager::LogEnemiesKilled(double count_enemies) {
```
paulbui081, line 52: Be sure to add a comment documenting what 'count_enemies' represents.

tests/tools/StatsTrackerTest.cpp
paulbui081, line 1: ### Testing and Error Checking
Looking through this unit test, it properly tests a majority of the public member functions created in the StatsTracker class except for DisplayData(...). They are well-organized and they properly put the functions through scenarios testing their intended purposes. 

The only small nitpicks I have is that there could be more test cases testing different edge cases or situations that could occur beyond the intended functionality. Leaving some comments as well, explaining the purpose of why you're passing in certain values would be helpful in following the test's logic here as well. Other than that, this looks solid to me!

JordanSebagh:
No warnings during compilation:
I didn't receive any warnings

Proper uses of namespaces
Most of the code is correctly inside namespace cse498. The main() test file is outside the namespace, which is fine, but it uses explicit cse498:: prefixes correctly instead of using namespace std or using namespace cse498.

Adheres to the C++ standard
The code mostly uses standard C++ safely. One concern is that ItemBase::GetOwnerID() relies on assert(IsOwned()), but if assertions are disabled, calling it when the item is not owned could still be unsafe. A safer design might return std::optional<size_t>.

Appropriate Modularity
The code is modular and easy to understand 
.
Uses modern C++ constructs
The code uses some modern C++ features well, including [[nodiscard]], std::make_shared, and std::make_unique. The analytics code also uses const references clearly.

Uses standard library algorithms
There is not much use of standard library algorithms in this specific section. Most of the code is direct method calls or simple conditionals, which is fine

Pointer handling
Pointer handling is mostly safe. The code avoids manual new and delete, and the test code uses std::make_shared. 

const correctness
Const correctness is good overall. Many getters are marked const, and parameters like const std::string& name and const WorldBase& world avoid unnecessary copies.

JordanSebagh:
Advanced C++
The code uses some advanced/modern C++ features, especially [[nodiscard]]. It does not show much use of templates, lambdas, ranges, or constexpr in this specific code, so that could be an area for improvement, although what you have here is fine how is.

source/Analyze/AnalyticsManager.cpp
```diff
@@ -11,15 +11,20 @@
...
 namespace cse498 {
 
-/*
-Resets the health log
-*/
-void AnalyticsManager::ResetHealthLog() { mHealthLog.Clear(); }
+/**
+ * Logs damage dealt in the current run and updates the total damage dealt.
+ */
+void AnalyticsManager::LogRunDamage(double amount) { mCurrentRunStats.damageDealt += amount; }
```
tmaks2, line 17: Maybe you should consider ignoring negative damage here similar to LogDamageDealt()

source/Agents/Classic/Enemy.cpp
```diff
@@ -33,4 +34,17 @@ std::size_t Enemy::ClaimGoldDrop() {
     mGoldClaimed = true;
     return mGoldDrop;
 }
+
+void Enemy::TakeDamage(double amount) {
+    AgentBase::TakeDamage(amount);
+    if (auto analytics = world.GetAnalyticsManager()) {
+        analytics->LogRunDamage(amount);
```
tmaks2, line 41: I think it's good that you are logging damage here. One thing I noticed though is that this only accounts for the damage amount passed in but not necessarily what was applied like in a case where the enemy has 5 HP left but takes 10 damage. Not sure if we want what was actually applied or just the damage taken total for analytics but might be something to consider

source/Interfaces/GUI/interface/Game.cpp
```diff
@@ -108,6 +111,18 @@ bool Game::Initialize() {
     if (!LoadCheck("player", std::string(ASSETS_DIR) + "Player/player.png"))
         return false;
 
+    // Analytics setup
+    mAnalyticsManager = std::make_shared<AnalyticsManager>();
+    mStatsTracker = std::make_unique<StatsTracker>();
+
+    // USED TO TEMPORARILY PUT IN VALUES
```
tmaks2, line 118: You might want to remove these now commented out temporary test values

PR #87, opened 2026-04-21
Group16

ncharters:
Final project code for Group 16. Additional code that we would like peer reviewed is also in PR #86

### DIFF
--- a/source/Analyze/AnalyticsManager.cpp
+++ b/source/Analyze/AnalyticsManager.cpp
@@ -46,10 +46,10 @@ void AnalyticsManager::Reset() {
     ResetActionLog();
 }
 
-
-/*
-Adds a new enemy count value to the enemies killed log
-*/
+/**
+ * Adds a new enemy count value to the enemies killed log
+ * @param count_enemies enemies
+ */
 void AnalyticsManager::LogEnemiesKilled(double count_enemies) {
     // Ignore negative values
     if (count_enemies < 0) {
@@ -70,9 +70,9 @@ void AnalyticsManager::LogDamageDealt(double damage) {
     mDamageDealtLog.Add(damage);
 }
 
-/*
-Returns a const reference to the enemies killed log
-*/
+/** 
+ * Returns a const reference to the enemies killed log
+ */
 const DataLog& AnalyticsManager::GetEnemiesKilledLog() const noexcept { return mEnemiesKilledLog; }
 
 /**
--- a/source/Analyze/AnalyticsManager.cpp
+++ b/source/Analyze/AnalyticsManager.cpp
@@ -5,25 +5,26 @@
  * The goal of this class is to act as a container to store all the DataLogs and
  * ActionLogs for the project. This class will be called by the game to log data
  * and actions which will be used by the StatsTracker class.
+ * Used AI to help create function comments
  **/
 
 #include "AnalyticsManager.hpp"
 
 namespace cse498 {
 
-/*
-Resets the health log
-*/
+/**
+ * Clears the health log
+ */
 void AnalyticsManager::ResetHealthLog() { mHealthLog.Clear(); }
 
-/*
-Resets the enemies tracked log
-*/
+/**
+ * Clears the enemies tracked log
+ */
 void AnalyticsManager::ResetEnemiesTrackedLog() { mEnemiesTrackedLog.Clear(); }
 
-/*
-Resets the damage dealt log
-*/
+/**
+ * Clears the damage dealt log
+ */
 void AnalyticsManager::ResetDamageDealtLog() { mDamageDealtLog.Clear(); }
 
 /**
@@ -41,9 +42,10 @@ void AnalyticsManager::Reset() {
     ResetActionLog();
 }
 
-/*
-Adds a new health value to the health log
-*/
+/**
+ * Adds a new health value to the health log
+ * @param health The health value to be added to the log
+ */
 void AnalyticsManager::LogHealth(double health) {
     // Ignore negative values
     if (health < 0) {
@@ -52,9 +54,10 @@ void AnalyticsManager::LogHealth(double health) {
     mHealthLog.Add(health);
 }
 
-/*
-Adds a new enemy count value to the enemies tracked log
-*/
+/**
+ * Adds a new enemy count value to the enemies tracked log
+ * @param count_enemies The enemy count value to be added to the log
+ */
 void AnalyticsManager::LogEnemiesTracked(double count_enemies) {
     // Ignore negative values
     if (count_enemies < 0) {
@@ -63,9 +66,10 @@ void AnalyticsManager::LogEnemiesTracked(double count_enemies) {
     mEnemiesTrackedLog.Add(count_enemies);
 }
 
-/*
-Adds a new damage dealt value to the damage dealt log
-*/
+/**
+ * Adds a new damage dealt value to the damage dealt log
+ * @param damage The damage dealt value to be added to the log
+ */
 void AnalyticsManager::LogDamageDealt(double damage) {
     // Ignore negative values
     if (damage < 0) {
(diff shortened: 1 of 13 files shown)


karensuzue:
**Reviewers for Group 16:** Please select all commits starting from **[44a1f8d](https://github.com/CSE498/Spring2026-CompanyA/pull/87/commits/44a1f8d627fadc66d33f05bd573d4a83cc74c55a) and earlier**

source/core/WorldBase.hpp
```diff
@@ -320,6 +320,27 @@ class WorldBase {
         return false;
     }
 
+    /**
+     * @brief Collects all actions from every agent's ActionLog and forwards them
+     *        into the AnalyticsManager's consolidated ActionLog.
+     * @note  Call this manually (e.g., from the GUI) whenever you want the
+     *        AnalyticsManager to reflect the latest agent activity.
+     */
+    void SyncAgentLogsToAnalytics() {
```
alexelkins23, line 329: I like that you put this function into WorldBase so it will be easy to access from both worlds and the GUIs. The design of the analytics being collected by each agent, then compiled by the world is nice, it keeps things simple by collecting data from the source and using the natural flow of the program to get the data where it needs to go.

alexelkins23 (commented):
There isn't a lot of new code in this PR, I think it is mostly in the statsMenu PR so I don't have much feedback to give on this one.

Code Design: I left a comment on the SyncAgentLogsToAnalytics function in WorldBase that I thought was good. That was pretty much the only code change in this PR, but looking at the other files the design of the module looks well thought-out and easy to implement in other modules.

Advanced C++: There aren't any new uses of advanced C++ features that I could find, but there also aren't really any good places to use them that I could find. The only thing I could think of is possibly templating DataSamples/DataLogs to accept any numeric type instead of just doubles, but I'm not sure it would be very useful.

source/Analyze/AnalyticsManager.cpp
paulbui081, line 1: ### Interoperability and API

Function names are very clear and the purpose of each public member function is easy to understand. The comments give further clarity into how the function and its parameters operate. Good stuff!

source/Analyze/AnalyticsManager.cpp
```diff
@@ -57,9 +58,10 @@ void AnalyticsManager::LogEnemiesKilled(double count_enemies) {
...
-/*
-Adds a new damage dealt value to the damage dealt log
-*/
+/**
+ * Adds a new damage dealt value to the damage dealt log
+ * @param damage The damage dealt value to be added to the log
+ */
 void AnalyticsManager::LogDamageDealt(double damage) {
     // Ignore negative values
     if (damage < 0) {
```
Lak453, line 67: Nice guard against invalid negative input here. It may also be worth adding a test to verify negative values are ignored as expected.

source/Analyze/AnalyticsManager.cpp
```diff
@@ -73,27 +75,34 @@ Returns a const reference to the enemies killed log
 */
 const DataLog& AnalyticsManager::GetEnemiesKilledLog() const noexcept { return mEnemiesKilledLog; }
 
-/*
-Returns a const reference to the damage dealt log
-*/
+/**
+ * Returns a const reference to the damage dealt log
```
Lak453, line 79: Since both the block comment and the Doxygen comment describe the same behavior, you could remove one of them to reduce repetition and make the file easier to scan.

source/tools/ActionLog.hpp
```diff
@@ -53,27 +60,39 @@ class ActionLog {
...
+     */
     int GetActionCountInRange(double startTime, double endTime) const;
 
-    /// Clears all actions and resets sequence counter (CurrentTime is preserved).
+    /**
+     * Clears all actions and resets sequence counter (CurrentTime is preserved).
+     */
     void Clear();
 
     /// add comments for params later gotta fix nahum issues first
```
JordanSebagh, line 98: need to add comments

source/tools/DataLog.cpp
```diff
@@ -94,9 +102,10 @@ std::optional<double> DataLog::Mean() const {
...
 
-/*
-Function returns the median of the values in the data log
-*/
+/**
+ * Function returns the median of the values in the data log
+ * @return The median of the values in the data log, or std::nullopt if the log is empty
+ */
 std::optional<double> DataLog::Median() const {
     if (mDataValues.empty()) {
```
JordanSebagh, line 110: you could use transform/ranges here instead

JordanSebagh (commented):
No warnings during compilation:
I recieved no warnings when I ran your code

Proper uses of namespaces
Most code is correctly placed inside namespace cse498. One minor consistency issue is that OutputManager.cpp uses cse498:: before each function instead of wrapping the whole file in namespace cse498 { }.

Adheres to the C++ standard
The code generally uses standard C++ features and avoids obvious undefined behavior. I did not see unsafe manual memory management. 

Appropriate Modularity
The code is well separated into classes with clear purposes. AnalyticsManager, StatsTracker, DataLog, ActionLog, OutputManager, and Timer each handle a specific responsibility.

Uses modern C++ constructs
The code uses modern C++ well. Good examples include std::optional, std::string_view, std::scoped_lock, lambdas, and constexpr.

Uses standard library algorithms
DataLog makes good use of algorithms like std::min_element, std::max_element, std::accumulate, and std::sort. One improvement would be using algorithms or ranges more in places like manually copying values for the median calculation.

Pointer handling
Pointer handling looks safe. I did not see manual new or delete, and the code mostly relies on standard library containers and value members.

const correctness
Const correctness is strong overall. Many getters are marked const, and functions that only read from objects take parameters as const references.

JordanSebagh:
Advanced C++
The code includes several advanced C++ features, especially std::optional, lambdas, constexpr, std::function, std::string_view, and std::scoped_lock. Templates and ranges are not used much, so that could be an area for improvement.

source/Analyze/StatsTracker.hpp
```diff
@@ -49,48 +52,65 @@ struct ActionSummary {
...
- * @details
- * Numeric summaries and action summaries are split so GUI code can render
- * them differently if needed.
- */
+ /**
+  * @struct DashboardSnapshot
+  * Combined dashboard payload produced by StatsTracker.
+  * Numeric summaries and action summaries are split so GUI code can render them differently if needed.
+  */
 struct DashboardSnapshot {
```
tmaks2, line 60: I like that you included a struct for interoperability since the GUI/WebUI groups can use simple summary structs instead of reading the raw logs directly

source/core/WorldBase.hpp
tmaks2, line 329: This seems like a great function for connecting the worlds/GUI to the analytics manager. One thing I'm not sure about is duplicate logging, since this goes through every agent's full ActionLog each time it's called, would calling it more than once add the same actions again?

source/tools/ActionLog.hpp
tmaks2, line 99: Definitely should add documentation for these functions

source/tools/DataLog.hpp
```diff
@@ -17,81 +17,99 @@
...
     double TimeOverThreshold(double threshold) const;
 
-    /*
-    Helper function whose purpose is to advance the Timer for timestamp testing purposes without manually waiting
-    */
+    /**
+     * Helper function whose purpose is to advance the Timer for timestamp testing purposes without manually waiting
+     * @param seconds The amount of time to advance by
+     */
     void advanceTimeForTesting(double seconds);
```
tmaks2, line 102: This should be PascalCase

tmaks2:
Overall, I think the analytics API is going in a good direction. I like the split between AnalyticsManager for storing/logging data and StatsTracker for creating summaries the GUI can use. You also did a great job of adding doxygen style documentation in this PR

The main thing I noticed from an API/interoperability standpoint are with SyncAgentLogsToAnalytics(). I may be misunderstanding part of the intended flow, but repeated calls seem like they could duplicate actions. I also think the demo/docs could explain the expected flow a little more clearly.


The other PR you referenced for review is also pretty strong in both code design and API/interoperability. The main things I noticed were small and had to do with logging damage taken. Overall, your module seems to be in a fairly good spot.

PR #88, opened 2026-04-21
Group15

GumdropHorse:
This is the group 15 submissions for the 4/20 peer review cycle.
In order to run our demo, follow these steps:
 - Run the following command from the root directory: cmake --build build --target group15_demo
 - cd into the demos folder
 - Run: ./group15_demo.exe
 
The main files to look at are the following:
source/Worlds/Dungeon/BSP-Dungeon.hpp
source/Worlds/Dungeon/CastleLevel.hpp
source/Worlds/Dungeon/CaveLevel.hpp
source/Worlds/Dungeon/DungeonBase.hpp
source/Worlds/Dungeon/DungeonWorld.hpp
source/Worlds/Dungeon/ForestLevel.hpp
source/Worlds/Dungeon/LevelBase.hpp
source/Worlds/Dungeon/RoomHolder.hpp
source/Worlds/Dungeon/WorldGeneration.hpp
tests/tools/GenerativeWorld/BSP-Dungeon.cpp

The main thing that we worked on were implementation with the GUI and agent groups, as well as finalizing the level generation process.


karensuzue:
**Reviewers for Group 15:** Please select all commits starting from **[07a6ff0](https://github.com/CSE498/Spring2026-CompanyA/pull/88/commits/07a6ff0d03694ea32d9017e6ff6d12eba09917c3) and earlier**

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,816 @@
...
+        size_t m_monster_tile_skeleton;
+        size_t m_monster_tile_goblin;
+        size_t m_secret_door_top;
+        size_t m_secret_door_bottom;
+        size_t m_secret_door_left;
+        size_t m_secret_door_right;
+        size_t m_exit_door_l1;
+        size_t m_exit_door_l2;
+        size_t m_exit_door_l3;
+        size_t m_exit_door_l4;
```
Plutolinus: m_exit_door_l4 is declared but never initialized, yet DoAction() still compares grid cell types against it. This is undefined behavior. In the worst case, it can randomly treat normal tiles as exits, or behave differently across compilers and optimization levels.

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,816 @@
...
+            auto bow3 = m_item_pool.Insert("Bow +3", 0.0);
+            auto bow4 = m_item_pool.Insert("Bow +4", 0.0);
+            auto bow5 = m_item_pool.Insert("Bow +5", 0.0);
+            auto healing_potion = m_item_pool.Insert("Healing Potion", 1.0);
+            auto defense_potion = m_item_pool.Insert("Defense Potion", 0.5);
+            auto speed_potion = m_item_pool.Insert("Speed Potion", 0.5);
+            auto axe = m_item_pool.Insert("Axe", 1.0);
+            auto pickaxe = m_item_pool.Insert("Pickaxe", 1.0);
+            auto shovel = m_item_pool.Insert("Shovel", 1.0);
+
```
Plutolinus, line 280: The combat logic depends on m_enemy_id, but in the constructor this ID is only reserved by calling GetNextAgentId() and is never actually associated with any spawned enemy. As a result, INTERACT is very unlikely to hit a real enemy. Worse, the code currently spawns PacingAgents, but the kill path dynamic_casts them to Enemy&; if the ID ever happens to match, this can throw or crash.

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,816 @@
...
+        size_t m_exit_door_l1;
+        size_t m_exit_door_l2;
+        size_t m_exit_door_l3;
+        size_t m_exit_door_l4;
+        //size_t m_variant_floor;
+
+        cse498::WeightedSet<std::string> m_item_pool; // A pool of possible items to generate
+
+        /// Provide the agent with movement actions.
+        void ConfigAgent(AgentBase &agent) override {
```
Plutolinus, line 179: The default demo input layer does not expose INTERACT, but DungeonWorld’s combat/interaction logic is entirely attached to WorldActions::INTERACT. So even if the enemy-ID issue above were fixed, the player still would not have a key available to trigger that path under TrashInterface.

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,816 @@
...
+                            *this
+                        );
+                        pickaxe->SetRange(BASE_RANGE);
+                        pickaxe->SetDamage(BASE_DAMAGE);
+                        pickaxe->SetHitBonus(BASE_BONUS);
+                        pickaxe->SetDropRate(1);
+                        pickaxe->SetHarvestSpeed(1);
+
+                        inventory.AddItem(std::move(pickaxe));
+                    } else if (item_str == "Shovel") {
```
Plutolinus, line 788: The "Shovel" branch actually constructs an ItemWeaponToolPickaxe, not an ItemWeaponToolShovel. This makes the API semantics inconsistent with runtime behavior, and it also suggests that the current long if/else item factory is prone to copy-paste regressions.

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -102,12 +134,58 @@ namespace cse498 {
...
+        void CreateBSPTree() {
+            insert_split(m_iterations); //Creates BSP Tree
+            PostOrderDFS(); //Grabs all the generated room slots from the tree
+        }
+
+        /**
+          * @brief Clears the BSP tree state for regeneration.
+          * @details Resets tree, leaf nodes, and exit door flag.
+          */
+        void ClearState() {
```
Plutolinus, line 151: BSP::ClearState() does not reset mExitProbabilityState. That means the probability of spawning the exit room on the second generation inherits state from the previous run, so regeneration and cross-level generation are no longer independent. It also makes seed-based tests less stable.

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Thank you for catching this! Fixed!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Thank you for catching this, fixed!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: Good Catch, fixed!

aryaplayzz:
Interoperability & API
The dungeon module plugs into WorldBase cleanly and exposes levels through a LevelBase abstraction, so other groups can construct and run a DungeonWorld without much friction. 

However, DungeonActions::MOVE_UP_STRING = "w" is published as the public action key while ConfigAgent registers it under "up", which is a silent contract mismatch. The combat path also depends on WorldActions::INTERACT, but TrashInterface does not expose that key, so the player cannot actually trigger combat from the demo. 

Several signatures lie about their intent — CreateDungeon(int&) never modifies its argument, GetDungeon() and GetBSP() return heavy objects by value, and GetLeafNodes()/GetBSPTree() return non-const references that let callers mutate internal state. 

The example driver in demos/group15_demo is good, and most public methods have Doxygen comments, but LoadLevelData()'s @return doc is stale, and the compile-time dependency on DUNGEON_ROOMS_DIR should be documented at the API boundary.

Code Design
The high-level decomposition (BSP → RoomHolder → WorldGeneration → DungeonWorld) is sensible, and ownership through unique_ptr is handled correctly.

The biggest issue is duplication: DoAction contains ~240 lines of nearly identical item-construction if/else branches, the three-level subclasses (ForestLevel, CaveLevel, CastleLevel) differ only in a directory string, and the 30+ mFloorIdL*V* fields plus 19-condition wall-blocking chain are all the same anti-pattern of data-as-code that an item factory, a parameterized Level class, and an unordered_set<size_t> mBlockingTileIds would each collapse cleanly. 

Const correctness is uneven (BSP::GetHeight/GetWidth aren't const, RoomHolder::GetCurrentRoom isn't const), dynamic_cast<Enemy&> on a PacingAgent is a runtime hazard, and mPlayerId is dead state. 

Modern C++ usage is otherwise solid, though .value() is unwrapped without has_value() checks in several spots, and PostOrderRoomConnect(BSPNode) takes its node by value, copying the room on every recursion.

source/Worlds/Dungeon/BSP-Dungeon.hpp
ZodiacalBee, line 15: #include <fstream>
#include <unordered_map>
#include <random>
#include <concepts>
#include <print>
are unused btw.

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -22,35 +22,67 @@
...
         int x, y, width, height; // (x,y) - (origin point of the grid (top-left corner))
         // (width, height) - dimension of the grid cut (lxw)
         std::string name; //Name of the node for debugging purposes
-        std::string file_name; //
+        std::string file_name; //placeholder just in case we're loading in images directly from BSPNodes
         std::vector<std::string> vector_room{}; //Stores a certain dungeon room
 
-
+        /// @brief Default constructor
         BSPNode() = default;
```
ZodiacalBee, line 45: it would be best to set x,y,width, height because otherwise they are uninitialized ints with garbage

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -22,35 +22,67 @@
...
+         * @param l Left child index
+         * @param r Right child index
+         * @param x X coordinate of top-left corner
+         * @param y Y coordinate of top-left corner
+         * @param width Width of the partition
+         * @param height Height of the partition
+         * @param name Debug name for the node
+         */
         BSPNode(int l, int r, int x, int y, int width, int height, std::string name)
             : left_child(l), right_child(r), x(x), y(y), width(width), height(height), name(name) {
```
ZodiacalBee, line 58: use std::move here no need to copy string twice

source/Worlds/Dungeon/RoomHolder.hpp
```diff
@@ -11,32 +11,42 @@
...
 namespace cse498 {
     ///Placeholders until we're able to define level states
     const int LEVEL_ONE = 1;
     const int LEVEL_TWO = 2;
     const int LEVEL_THREE = 3;
+    const std::string EXIT_FILE_STRING = "exit.txt";
 
+
+    /**
+     * @class RoomHolder
```
ZodiacalBee, line 28: minor but this docstring is modifying the wrong class. 
also cse498:: below can be removed, also no need to do m_rng() in constructor as default constructor is called automatically

source/Worlds/Dungeon/CastleLevel.hpp
```diff
@@ -0,0 +1,109 @@
...
+         * @details
+         * Iterates through ROOM_DATA and inserts each room into a
+         * WeightedSet using its corresponding file path and weight.
+         *
+         * @return A WeightedSet containing room file paths and weights.
+         *
+         * @note Each insertion is expected to succeed. Assertions are used
+         * to enforce this during development.
+         */
+        static cse498::WeightedSet<int> MakeRoomPool() {
```
ZodiacalBee, line 68: again ctrl + f for cse498:: and remove all

source/Worlds/Dungeon/CastleLevel.hpp
```diff
@@ -0,0 +1,109 @@
...
+            cse498::WeightedSet<int> rooms;
+
+            for (const auto &[num, weight]: ROOM_DATA) {
+                auto result = rooms.Insert(num, weight);
```
ZodiacalBee, line 72: technically a narrowing conversion of `num` from size_t --> int. I get IDE warning, no reason not to do static_cast<int> since this will be done implicitly, and it removes the warning

source/Worlds/Dungeon/ForestLevel.hpp
ZodiacalBee, line 1: Could shorten this all by moving variables to LevelBase as protected and defining private static functions there instead of repeat code between the levels. This applies to all the levels that are copy paste of this, but I'm putting this comment on one file

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+#include "../../core/item/ItemWeaponToolAxe.hpp"
+#include "../../core/item/ItemWeaponToolPickaxe.hpp"
+#include "../../core/item/ItemWeaponToolShovel.hpp"
+#include "../../Agents/Classic/PlayerFeatures/Inventory.hpp"
+
+namespace cse498 {
+
+    struct DungeonActions // just so they aren't globals
+    {
+        // Note 0 should ALWAYS Be stay based on AgentBase GetActionID implementation
```
ZodiacalBee, line 39: why is this copy pasted and renamed? I'd add a comment for clarity

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+        static constexpr size_t MAX_ID = 1000000000;
+
+    private:
+        int mLevelNum = 1; //Current int value level that the game is on
+        std::unique_ptr<LevelBase> mLevel = std::make_unique<ForestLevel>();
+
+        //The currently pointed to level that the player agent is on
+		std::vector<size_t> mSpawnedEnemyIds;
+
+        size_t mPlayerId;
```
ZodiacalBee, line 72: if this was copied from group2 demo then remove it. Ids shouldn't be stored in the world especially since the player can always be grabbed and then the id gotten. Keep one "single source of truth" which is the player itself that has this value. 

there are multiple enemies so that value isn't worth much

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+        size_t mEnemyId;
+
+		//Track level change to prevent an extra enemy turn
+		bool mLevelJustAdvanced = false;
+        /**
+         * @brief Builds the room pool for the current level.
+         * @return WeightedSet of room file paths and weights.
+         */
+        void LoadLevelData() {
+            std::cout << "Currently on level: " << mLevelNum << std::endl;
```
ZodiacalBee, line 82: side note: it would be nicer if this cout was on a debug only flag since otherwise code tests and gameplay would just spam this to console for no purpose

source/Worlds/Dungeon/BSP-Dungeon.hpp
ZodiacalBee, line 278: these should be const no?

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -167,6 +227,10 @@ namespace cse498 {
                     }
                 }
             }
+
+            for (auto i: grid) {
```
ZodiacalBee, line 231: don't copy. const auto i&

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+#include "../../tools/WeightedSet.hpp"
+#include "ForestLevel.hpp"
+#include "CaveLevel.hpp"
+#include "CastleLevel.hpp"
+#include "LevelBase.hpp"
+#include "../../Interfaces/TrashInterface.hpp"
+#include "../../Agents/PacingAgent.hpp"
+#include "../../Agents/Classic/Enemy.hpp"
+
+#include "../../core/item/ItemWeaponSword.hpp"
```
Deni-Tepic, line 25: it might be cleaner to centralize that logic in a factory or manager class

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+        // Note 0 should ALWAYS Be stay based on AgentBase GetActionID implementation
+        static constexpr std::size_t REMAIN_STILL = 0; // "stay"
+        static constexpr std::size_t MOVE_UP = 1; // "w"
+        static constexpr std::size_t MOVE_DOWN = 2; // "s"
+        static constexpr std::size_t MOVE_LEFT = 3; // "a"
+        static constexpr std::size_t MOVE_RIGHT = 4; // "d"
+        static constexpr std::size_t INTERACT = 5; // "e"
+        static constexpr std::size_t QUIT = 6; // "q"
+
+        static constexpr std::string REMAIN_STILL_STRING = "stay";
```
Deni-Tepic, line 48: Since these are compile-time constants, using constexpr const char* or std::string_view may be more appropriate than std::string

source/Worlds/Dungeon/LevelBase.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "../../tools/WeightedSet.hpp"
+
+namespace cse498 {
+    /**
+     * @class LevelBase
+     * @brief Abstract base class defining the interface for dungeon level configurations.
+     *
+     * @details Derived classes must implement methods to provide the room pool
+     */
+    class LevelBase {
```
Deni-Tepic, line 21: This base class is clean, but currently provides only minimal shared interface (room pool and directory). 

Since multiple level types exist, it may be worth considering whether more shared behavior could be moved into this base class to reduce duplication across derived level implementations

source/Worlds/Dungeon/LevelBase.hpp
```diff
@@ -0,0 +1,32 @@
...
+    public:
+        virtual ~LevelBase() = default;
+
+        [[nodiscard]] virtual const cse498::WeightedSet<int> &GetRoomPool() const = 0;
+
+        [[nodiscard]] virtual const std::string &GetRoomDir() const = 0;
```
Deni-Tepic, line 27: Returning a const std::string& assumes the underlying string has stable lifetime. If this is always backed by a member variable in derived classes this is fine, but otherwise returning by value or using std::string_view could be safer

source/Worlds/Dungeon/ForestLevel.hpp
```diff
@@ -0,0 +1,107 @@
...
+		 * @details
+		 * Initializes the base ForestLevel with a generated room pool
+		 */
+		ForestLevel()
+		: m_room_pool(MakeRoomPool()) {}
+
+		/**
+		 * @brief Default destructor.
+		 */
+		~ForestLevel() = default;
```
Deni-Tepic, line 93: Since the base class already has a virtual destructor,
this destructor could be marked override for clarity or omitted entirely if no custom cleanup is needed

source/Worlds/Dungeon/BSP-Dungeon.hpp
Deni-Tepic, line 1: the function takes BSPNode by value, which may result in unnecessary copying during recursion. Passing by reference would likely improve performance and avoid repeated copies

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -80,16 +135,35 @@ namespace cse498 {
...
+		std::vector<std::pair<int, int>> m_goblin_spawns;
+
+
+    	/// @brief Locations where skeletons will spawn
+    	std::vector<std::pair<int, int>> m_skeleton_spawns;
+
         /// @brief takes the leaf node's (x,y) coordinates and room information from the BSP_Tree and translates it onto mGrid
         /// @param node BSPNode filled with room information (x/y coords, room width/height, room vector string)
         void RasterizeGrid(const BSPNode &node) {
             int room_height = node.vector_room.size();
```
ZodiacalBee, line 162: these are all narrowing conversions from int --> size_t

Deni-Tepic (commented):
The overall structure makes sense and is easy to follow, especially how the different parts of the dungeon system are separated. There is some repeated code and a few hardcoded values that could maybe be simplified, but nothing major. 
On the C++ side it looks solid overall, with just a few small things like const usage and avoiding extra copies that could be improved. Overall, this is a good implementation with a strong foundation.

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -105,6 +179,14 @@ namespace cse498 {
                     char c = node.vector_room[y][x];
 
                     //if (c == '#') continue; //Skips the outer outline of the room
+					// Find goblin spawn locations
+					if (c == 'g') {
+						m_goblin_spawns.push_back({grid_x, grid_y});
```
ZodiacalBee, line 184: Clion IDE highlights this for me saying to use emplace_back instead so that construction happens directly in the vector rather than creation of the object then movement into the vector.

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -116,6 +198,8 @@ namespace cse498 {
         /// @param room we're inputting the BSP_Tree Node's room value
         /// @return pair of coordinates
         [[nodiscard]] Point CalcRoomCenter(const std::vector<std::string> &room) const {
```
ZodiacalBee, line 200: can be static. also do static_cast int on the width and height for the returns to make it explicit. It is a little ambiguous readability-wise that it is truncated to an int

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -125,17 +209,21 @@ namespace cse498 {
...
+         * @brief Connects the Rooms stored within the BSP Nodes taking into consideration their ascii symbol
+         * 
+         * @details Calculates the distance (x-y) between two rooms) and connects a corridor tunnel between them 
+         * 
          */
-        void ConnectTunnel(LinkedRooms RoomCoordinates) {
+        void ConnectBSPRooms(LinkedRooms RoomCoordinates) {
             auto [x1_value, y1_value, x2_value , y2_value] = RoomCoordinates;
-            auto const wall_set = {'^', '>', '<', '&'};
+            auto const wall_set = {'?', '!', '$', '@',
```
ZodiacalBee, line 224: picky but I find it bad practice to define this here in the middle of the file. Preferably it is defined inline static constexpr std::array WALL_SET{ ...)

source/Worlds/Dungeon/WorldGeneration.hpp
ZodiacalBee, line 234:
```
bool negative_y = point_y < 0;
bool negative_x = point_x < 0;

point_y = std::abs(point_y);
point_x = std::abs(point_x);
```

shorter, cleaner, no branching

source/Worlds/Dungeon/WorldGeneration.hpp
```diff
@@ -174,15 +262,15 @@ namespace cse498 {
...
         }
 
-        /// @brief DFS to go through the populated BSP tree in order to connect rooms together
+        /// @brief Post-order DFS algorithm that searches BSP Tree
+        /// @details DFS to go through the populated BSP tree in order to connect rooms together
         /// @param node BSPNode filled with room information (x/y coords, room width/height, room vector string)
         /// @return (x,y) pair coordinate struct of room's location in the grid
-        Point PostOrderRoomConnect(BSPNode node) {
-            //Need to change Node node to idx int later
+        Point PostOrderRoomConnect(BSPNode node) {
```
ZodiacalBee, line 273: const BSPNode& ?

source/Worlds/Dungeon/BSP-Dungeon.hpp
```diff
@@ -102,12 +134,59 @@ namespace cse498 {
...
+        void RoomCompatibility() {
+            // bool valid_room = false; // To ensure that a valid room is found when generating dungeon
+
+            auto &lower_threshold = mExitProbabilityState; //lower prob bound
+            const auto upper_threshold = DOOR_EXIT_PROBABILITY_LIMIT; //upper prob bound
+            const auto probability_increment = DOOR_EXIT_PROBABILITY_INCREMENT;
+
+            if (!m_exit_door) {
+                auto val_one = m_rng.GetValue(0.0, 1.0);
+                if ((val_one.value() < lower_threshold && !m_exit_door)) {
```
ZodiacalBee, line 169: redundant (!m_exit_door) check. It is always true

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
```diff
@@ -78,23 +80,25 @@ TEST_CASE("BSP-Dungeon Constructor", "[core]") {
...
 
     SECTION("Testing that RegenerateObjectState() regenerates BSPTree properly") {
-        cse498::BSP BSP(MakeRoomPool(), SEED_VALUE_ONE, FILE_PATH);
+        cse498::ForestLevel level;
+        cse498::BSP BSP(level, SEED_VALUE_ONE);
         SetDefaultSetting(BSP);
 
         BSP.ClearState();
 
         CHECK(BSP.GetBSPTree().size() == 0);
```
ZodiacalBee, line 89: could do .Empty for readability

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
ZodiacalBee, line 1: Following directions i have to note all public member functions not used in this test file:
insert_split -- though tested through CreateBSPTree so passable. 
RoomCompatibility & PostOrderDFS -- but both are tested through constructor

GenerateTileMap - though you define it as a debug function.


I think you are fine.

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+        //     }
+        //     return;
+        // }
+
+		void RunAgents() override {
+			TrashInterface* ui = nullptr;
+
+			for (const auto& agent_ptr : agent_set) {
+				if (agent_ptr->IsInterface()) {
+					ui = static_cast<TrashInterface*>(agent_ptr.get());
```
ZodiacalBee, line 389: dynamic_casts for conversions between polymorphic structures

this is a weird way to get the UI. I really would rather have a simple function and storage in the world for the UI pointer and in the game/ui when the world is created just call world.SetUI(ui) just leaving the assert in here

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+                    if (dx <= 1.0 && dy <= 1.0) {
+                        interacted = true;
+                        if (other.GetID() == mEnemyId) {
+                            const double dealt = DamageCalculator::Calculate(GetPlayer()->GetStats(), GetAgent(mEnemyId).GetStats());
+                            other.TakeDamage(dealt);
+                            std::cout << agent.GetName() << " hits enemy for " << static_cast<int>(dealt) << " damage.\n";
+                            if (!other.IsAlive()) {
+                                auto& enemy = dynamic_cast<Enemy&>(other);
+                                auto& player = dynamic_cast<PlayerAgent&>(agent);
+                                HandleEnemyDefeat(enemy, player);
```
ZodiacalBee, line 486: it is late in this project which changes things but I could make interct return a reference to the thing interacted with which would allow calling handle enemy defeat. I figure this would've been called on cleanup dead agents or whatever it is called though --> which I imagine would be called after processing through all the agents

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+				// Currently there is no level 4
+				// || main_grid[new_position] == mUpperExternalL4
+                // || main_grid[new_position] == mLowerExternalL4
+                // || main_grid[new_position] == mLeftExternalL4
+                // || main_grid[new_position] == mRightExternalL4
+                // || main_grid[new_position] == mInternalObstacleL4V1
+                // || main_grid[new_position] == mInternalObstacleL4V2
+			) { return false; }
+
+            if (main_grid[new_position] == mLootTile && agent.GetID() == 0) {
```
ZodiacalBee, line 557: I regret not encouraging separation of player logic from other agents. This is what I would've done too

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+                        inventory.AddItem(std::move(sword));
+                    } else if (item_str == "Sword +1") {
+                        std::unique_ptr<ItemWeaponSword> sword1 = std::make_unique<ItemWeaponSword>(
+                            item_id,
+                            item_str,
+                            "/assets/item/item_sword_1.png",
+                            BASE_GOLD + 2,
+                            *this
+                        );
+                        sword1->SetRange(BASE_RANGE);
```
ZodiacalBee, line 585: pretty sure my team worked on setting default values for the weapons. It may conflict but that is okay I guess. I understand

source/Worlds/Dungeon/DungeonWorld.hpp
```diff
@@ -0,0 +1,833 @@
...
+                // || main_grid[new_position] == mInternalObstacleL4V1
+                // || main_grid[new_position] == mInternalObstacleL4V2
+			) { return false; }
+
+            if (main_grid[new_position] == mLootTile && agent.GetID() == 0) {
+                if (agent.IsPlayerAgent()) {
+                    Inventory &inventory = agent.GetInventory();
+                    std::string item_str = GetRandomItem();
+                    Random rng;
+                    size_t item_id = rng.GetValue(MIN_ID, MAX_ID).value();
```
ZodiacalBee, line 562: wait I think this causes issues. If the item is unique then the id should be unique to identify it otherwise it will be impossible to know what it is. Things may still work but I'd rather have some size_t GetNextItemId type thing or something.... I thought we had something like that already. I don't recall what my group did for this issue in trading

source/Worlds/Dungeon/RoomHolder.hpp
```diff
@@ -52,6 +62,11 @@ namespace cse498 {
             m_current_room = determined_room;
         }
 
+        ///@brief Overloaded function dedicated to loading in the exit room to the next level, separate from the room_pool
+        void SetCurrentRoom(std::vector<std::string> exit_room) {
+            m_current_room = exit_room;
```
ZodiacalBee, line 67: do std::move() here. I don't think it does anything automatically or do const &

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
yhenry6, line 1: Testing & Error Checking

Unit Testing
Good job testing the constructor, setters/getters, ClearState(), and regeneration behavior. A small thing to improve is deeper validation of the generated tree. 

Edge Cases
Overall, it looks like most scenarios are handled. You test multiple valid settings and seeds, which gives the BSP generation some useful variety. If you wanted to make the tests stronger, you could add a few more boundary-style cases, like very small maps or settings where splitting is limited.

Interoperability and API
The test file shows that the BSP class can be created with a ForestLevel, which is good for integration with other modules.

source/Worlds/Dungeon/BSP-Dungeon.hpp
yhenry6, line 1: Testing & Error Checking
Unit Testing
The header gives a good amount of public functionality that can be tested, like construction, regeneration, tree creation, setters/getters, and seed control. 

Edge Cases
The code handles some edge cases well, such as stopping the split when a node is too small to divide further. It could still use clearer validation for unusual settings like very small dimensions, invalid width/height, or negative iteration values.

Assertions
Good use of assert in PostOrderDFS() to make sure the BSP tree is initialized before traversal, which is a good invariant check. It may also be worth asserting or validating dimensions, iterations, and node bounds.

Error Handling
The code has good internal error handling with std::optional in random_split(), which lets the tree stop safely when splitting is not possible. 

Interoperability and API
The class is documented well and has useful public methods for other groups to create, regenerate, and inspect BSP trees. One small issue is GetBSPTree() and GetLeafNodes() returning non-const references, which allows outside code to modify internal state directly.

paulbui081: you're right, fixed!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: good catch, fixed!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: fixed!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: fixed!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: sounds good, added!

source/Worlds/Dungeon/BSP-Dungeon.hpp
paulbui081: good catch, fixed!

source/Worlds/Dungeon/WorldGeneration.hpp
paulbui081: good idea, I'll move it up as a named constant

source/Worlds/Dungeon/WorldGeneration.hpp
paulbui081: very nice, I'll substitute this in. Thanks!

source/Worlds/Dungeon/WorldGeneration.hpp
paulbui081: fixed!

source/Worlds/Dungeon/RoomHolder.hpp
GumdropHorse: Thanks for catching this, fixed!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Thanks for the feedback! Resolved!

source/Worlds/Dungeon/WorldGeneration.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/WorldGeneration.hpp
GumdropHorse: The reason it is not static, is because the function is const. However, thanks for catching the thing about width and height! That has been resolved.

tests/tools/GenerativeWorld/BSP-Dungeon.cpp
GumdropHorse: Good idea, changed!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Thank you for this, resolved!

source/Worlds/Dungeon/CastleLevel.hpp
joeyhyun1: done

source/Worlds/Dungeon/WorldGeneration.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Good idea, just got that implemented!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: I moved it here to resolve issues with other class having access to it, and to allow us to customize the actions to our world if needed. Just added a comment for clarity, thanks!

source/Worlds/Dungeon/CastleLevel.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/RoomHolder.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/LevelBase.hpp
GumdropHorse: Thanks for pointing this out! However, the derived functions always use a member variable to back this.

source/Worlds/Dungeon/ForestLevel.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/DungeonWorld.hpp
GumdropHorse: Thanks for catching this, resolved!

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: I'm not sure what is meant by this? The referenced section does not have dynamic casting. If you are saying we should have used dynamic casting instead of static cast I disagree. Based on oop principles its bad practice to dynamic cast. Also, this is not a conversion, it is grabbing the interface and storing it within a variable for later use

source/Worlds/Dungeon/LevelBase.hpp
AbbyFranz: We agree with you that shared behavior should be moved into the base class :)

For the sake of time and worrying about full company integration, we will probably leave this as is for now. If we had longer we'd definitely change this

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: I agree. If we had more time before the deadline we would definitely make that change

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: Yes, if we had more time i'm sure there is probably a better way, but for the time being it'll have to do. Also handleEnemyDefeat may just be a confusing name, since I think it is just handling exp collection from a dead enemy. The clean up for dead enemies does have a seperate function that is called elsewhere (I think on each turn and on a new level load)

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: Thanks!

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: Alot of this code was pulled in based on group 2 code... If that is your team lol. If there are conflicts pls msg us on discord!

source/Worlds/Dungeon/DungeonWorld.hpp
AbbyFranz: This is not copied from group 2. mSpawnedEnemyIds helps with tracking/cleaning up spawned enemies on each level load. Since all levels are contained within a single world we needed a special way to track the ones on the current level and then make sure they are all cleaned up before we generate in new ones on the next level

source/Worlds/Dungeon/ForestLevel.hpp
AbbyFranz: Yes, we agree this is a good cleanup. Some of the stuff seems duplicated (like the weighted sets) but if some dungeons had more rooms or had more unique probabablities for spawning rooms it would have to be within the individual class anyways.

PR #91, opened 2026-04-21
Group14

lunncurt:
The ```docs/group14.md``` gives an accurate state of the module including current relevant classes to review (or at least main contributors to current game state).

Main changes this submission were:
```
  - demos/InteractiveWorld/Group14_main.cpp
  - source/Worlds/Hub/InteractiveWorld.hpp
  - source/Agents/AI/FetchAgent.hpp
  - source/Agents/Classic/ResourceManagementAgent.hpp/cpp
  - source/Worlds/Hub/Building.hpp
  - source/Worlds/Hub/ResourceProducer.hpp
```
  After that, these are the main supporting files:
```
  - source/Worlds/Hub/ResourceSpawn.hpp
  - source/Worlds/Hub/ResourceBank.hpp
  - source/Worlds/Hub/TownHall.hpp
  - source/Interfaces/TrashInterface.hpp
  - source/Worlds/Hub/InteractiveWorldSaveManager.hpp
```


karensuzue:
**Reviewers for Group 14:** Please select all commits starting from **[0542cc9](https://github.com/CSE498/Spring2026-CompanyA/pull/91/commits/0542cc9087c8d1d712b2658150fbcdb68ddf7f83) and earlier**

AB123456kkkkk (commented):
Testing and Error Checking
TestFarmingAgent — Overall there are unit tests for every single public member function (including constructors). The test cases are also well organized.  Most of the private portions from the header and cpp files are included in the code. One edge case that is tested is null pointers.
TestFetchAgent — Unit tests for all public member functions (including constructors). The test cases are also well organized.  Most of the private portions from the header and cpp files are included in the code. One edge case that is tested out of bounds. No asserts in this file and TestFarmingAgent.
ResourceBank — Unit tests for all public member functions (including constructors). The test cases are also well organized.  Most of the private portions from the header and cpp files are included in the code. Edge case include boundary conditions.
TestResourceManagementAgent — This test case is different than the others one, mainly due to more checks in the file.
Overall I liked the fact that there is good unit testing in the code. All of the test cases are well organized and trigger cases that exercise the private portions. All of the edge cases have been tested. Error conditions are checked well. Good error handling. 


Implementation
All of the code is under the cse498 namespace, which is good practice. The files do not have any namespace pollution and no use of using namespace std.  No unnecessary code duplication is there, appropriate modularity is maintained. Great use of memory management — used shared pointers and unique pointers throughout all of the files and no use of raw pointers. That is one of things that I liked about the code is the use of memory management. Const correctness is excellent throughout the files. Member functions are marked as const and parameters are also marked as const. For modern C++ constructs — there are uses of lambdas (2). Used templates. Multiple uses of std::optional, no uses of constexpr. Mixed both range-based for loops and standard uses. I was able to run the code perfectly — no errors or warnings during compilation. The code avoids undefined behavior. Used nodiscard whenever applicable. There are uses of algorithms.

demos/InteractiveWorld/Group14_main.cpp
```diff
@@ -5,14 +5,17 @@
  **/
```
murra381, line 6: Code Design: Clear enough to understand. The configureFetcher lambda improves readability. I thought that was good.

Advanced C++: Uses std::make_shared, std::make_unique, and a lambda well.

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -5,6 +5,7 @@
  **/
```
murra381, line 6: Code Design: Interface is understandable, I thought AddBuilding, AddResourceSpawn, AddResourceBank, and AddTownHall were easy to read and simple. TryInteractAdjacent is readable. I liked the small change in PrintInventory as well.

Advanced C++: Good use of std::array for interaction priority and std::shared_ptr/std::make_shared for inventory.

source/Agents/AI/FetchAgent.hpp
```diff
@@ -0,0 +1,331 @@
+/**
+ * This file is for the Fall 2026 CSE 498 section 2 Capstone project.
+ * @brief Agent that shuttles between an origin point and a deposit point.
+ */
```
murra381, line 4: Code Design: Strong interface overall. My main concern is the raw pointer ownership in m_origin and m_depositPoint; the code appears to assume those endpoint agents outlive the FetchAgent. I’d like to see that lifetime contract documented more clearly or represented more safely if possible.

Advanced C++: Good use of std::function, lambdas, [[nodiscard]], and structured default behavior. The BFS implementation is solid. No major notes here. Good job.

source/Agents/Classic/ResourceManagementAgent.cpp
```diff
@@ -0,0 +1,548 @@
+/**
+ * @file ResourceManagementAgent.cpp
+ * @brief Interactive world agent for upgrading buildings and selling resources.
```
murra381, line 3: Code Design: The public API is functional and easy to follow, with clear operations for upgrading buildings, selling resources, hiring lanes, and querying lane state. The upgrade, sell, and hire logic is separated into methods that can be tested directly, which is a strong design choice. Good job.

Advanced C++: Good use of std::optional, std::string* optional output messages, and aliases like GoldAmount. The parsing helpers are a nice small modular choice too. Good job with that.

source/Worlds/Hub/Building.hpp
```diff
@@ -7,6 +7,7 @@
 #pragma once
```
murra381, line 8: Code Design: This was good. Had upgradeable building plus local resource storage. Good use of private validation helper.  m_rateModifier = 0.25 could probably be a named static constexpr.

Advanced C++: Uses std::expected, std::optional, constexpr, and value semantics for BuildingUpgrade. This was good.

source/Worlds/Hub/ResourceProducer.hpp
```diff
@@ -10,7 +10,7 @@
 #include <memory>
 
 #include "Building.hpp"
```
murra381, line 12: Code Design: Small nitpick but m_outputType seems to be stored and never used after creation. It also stores raw pointers to Building and ResourceSpawn. Could maybe use references or non-owning pointers, but that is another nitpick.

Advanced C++: Uses std::chrono well. Could use constexpr defaults or stronger types for production rates, but current use is reasonable.

source/Worlds/Hub/ResourceSpawn.hpp
```diff
@@ -0,0 +1,82 @@
+/**
+ * This file is for the Fall 2026 CSE 498 section 2 Capstone project.
```
murra381, line 2: Code Design: The interface is simple and easy to understand, but I would add validation for negative values. AddResource() should reject negative quantities, and SetMaxCollectionQuantity() should reject negative limits. Otherwise Collect() can return a negative amount or accidentally increase the stored resource count. What exactly is stopping me from inputting -5 to AddResource()?

Advanced C++: Could benefit from static constexpr int DEFAULT_MAX_COLLECTION_QUANTITY = 10.

source/Worlds/Hub/ResourceSpawn.hpp
```diff
@@ -0,0 +1,82 @@
...
+ * @brief Represents a spawn point for resources that the Agents will grab from
+ * @note Status: PROPOSAL
```
murra381, line 4: ResourceBank.hpp
Code Design: I thought the code design for this class was clean.

Advanced C++: Uses a type alias from InteractiveWorldInventory, [[nodiscard]], and composition cleanly. Good design.

source/Worlds/Hub/TownHall.hpp
```diff
@@ -0,0 +1,47 @@
+/**
+ * This file is part of the Fall 2026, CSE 498, section 2, course project.
```
murra381, line 2: Code Design: This one needs some work. The class comment says @class Building, which is incorrect. The constructor that leaves m_inventory null makes DepositResource() unsafe. If someone creates a TownHall using the constructor without an inventory, then m_inventory is nullptr. Calling DepositResource() after that would dereference a null pointer, which can crash or cause undefined behavior.

Advanced C++: Uses std::shared_ptr and std::map.

source/Worlds/Hub/InteractiveWorldSaveManager.hpp
```diff
@@ -4,6 +4,7 @@
 #include <string>
 
 #include "../../../third-party/json/json.hpp"
```
murra381, line 6: Code Design: Useful single-purpose class. The save and load responsibilities are easy to understand, and the code is organized clearly around inventory, buildings, and resource-manager state. I thought this class was good overall.

Advanced C++: Uses nlohmann::json cleanly and makes good use of range-based loops. The implementation is straightforward and readable.

murra381 (commented):
Sorry this is a little late.

source/Worlds/Hub/InteractiveWorld.hpp
finfroc3: TODO: API??

source/Interfaces/TrashInterface.hpp
finfroc3: TODO: API??
is this one needed even

tests/Agents/TestResourceManagementAgent.cpp
finfroc3: TODO: test??

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -23,12 +26,13 @@ namespace cse498 {
 class InteractiveWorld : public WorldBase {
 protected:
     // World Inventory
-    InteractiveWorldInventory m_inventory;
+    std::shared_ptr<InteractiveWorldInventory> m_inventory = std::make_shared<InteractiveWorldInventory>();
```
marksandychez, line 29: Good use of a shared_ptr when an object will be referenced by multiple classes

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -60,59 +91,83 @@ class InteractiveWorld : public WorldBase {
...
     void PrintInventory() {
         std::ostringstream output;
-        output << m_inventory.GetAmount(ItemType::Wood) << ' ' << ItemTypeToString(ItemType::Wood) << " | "
-               << m_inventory.GetAmount(ItemType::Stone) << ' ' << ItemTypeToString(ItemType::Stone) << " | "
-               << m_inventory.GetAmount(ItemType::Metal) << ' ' << ItemTypeToString(ItemType::Metal);
+        output << m_inventory->GetAmount(ItemType::Wood) << ' ' << ItemTypeToString(ItemType::Wood) << " | "
+               << m_inventory->GetAmount(ItemType::Stone) << ' ' << ItemTypeToString(ItemType::Stone) << " | "
+               << m_inventory->GetAmount(ItemType::Metal) << ' ' << ItemTypeToString(ItemType::Metal);
 
         std::cout << output.str() << std::endl;
```
marksandychez, line 98: Could potentially upgrade to use print for more modern C++ usage, though this has zero effect on actual code execution.

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -60,59 +91,83 @@ class InteractiveWorld : public WorldBase {
...
 
+
 public:
     /**
      * Constructor
      */
     InteractiveWorld() {
         floor_id = main_grid.AddCellType("floor", "Floor that agents can walk on.", ' ');
         wall_id = main_grid.AddCellType("wall", "Impenetrable wall.", '#');
+        building_id = main_grid.AddCellType("building", "An impassable building.", 'B');
```
marksandychez, line 126: "Floor" and "Wall" could cause issues with the GUI team, though this can be further discussed with them. Image assets require unique "name" implementation to be drawn.

source/Agents/AI/FetchAgent.hpp
```diff
@@ -0,0 +1,306 @@
...
+            if (it->second == from) {
+                break;
+            }
+            step = it->second;
+        }
+
+        if (step == from.Up()) {
+            return GetActionID("up");
+        }
+        if (step == from.Down()) {
```
marksandychez, line 293: Could this be shorted to an if-elif structure?

source/Agents/Classic/FarmingAgent.cpp
```diff
@@ -1,8 +1,242 @@
...
+        {
+            // building level can scale how much stock farmer replenishes
+            amount += mAssignedBuilding->GetCurrentLevel();
+        }
+
+        RestockLimitedOffer(mRestockItemName, amount);
+
+        // Reset work timer and start heading home after work
+        mTicksSinceWork = 0;
+        mState = FarmerState::ReturningHome;
```
marksandychez, line 176: Good use of an enumeration

source/Agents/Classic/FarmingAgent.cpp
```diff
@@ -1,8 +1,242 @@
...
+    }
+
+    std::size_t FarmingAgent::SelectAction(const WorldGrid& grid)
+    {
+        ++mTicksSinceWork;
+
+        if (mAssignedBuilding == nullptr)
```
marksandychez, line 183: I believe the Google Style Guide usually prefers implicit pointer checks like:
`if (ptr) { }`
But this does not effect code

source/Worlds/Hub/ResourceSpawn.hpp
```diff
@@ -0,0 +1,82 @@
...
+     */
+    size_t SelectAction(const WorldGrid&) override { return 0; }
+
+    /**
+     * Add resources to the pool
+     * @param quantity quantity to add
+     */
+    void AddResource(const int& quantity) {
+        // Check for integer overflow
+        if (quantity > std::numeric_limits<int>::max() - m_quantity)
```
marksandychez, line 52: Great edge case checking, though mathematically I'd be impressed if this was hit. Still great standard practice

source/Worlds/Hub/Building.hpp
```diff
@@ -99,6 +103,18 @@ class Building {
      * @return current level as int
      */
     int GetCurrentLevel() const { return m_level; }
+    /**
+     * Set the current building level
+     * @param level level to set building to
+     * @return bool if setting level was successful
+     */
+    [[maybe_unused]] bool SetCurrentLevel(int level) {
```
marksandychez, line 111: This feels like something that maybe shouldn't have it's return value ignored, and could be changed to [[nodiscard]]. If there is a reason, I might just be missing it

marksandychez (commented):
# Advanced C++
A large amount of these classes revolve around integer calculation and proper resource transitions, which don't leave much room for overly many advanced C++ features. Despite this, I would consider the chrono clock to resource generation an example of a well implemented advanced feature. There are also a variety of shared_ptrs in cases where Buildings will need to refer to a common inventory object. The code also makes great use attributes to indicate the relevance of the return values of the functions. There might be potential for use of a lambda when checking the results of up, down, left, and right returns, but is not truly needed for the simple boolean checks.

# Implementation Quality

- **No Warnings During Compilation**
With the most recent changes compiled and merged onto group 17s branch, there were no error stemming from any code.

- **Proper Use of Namespaces**
Files properly indicate cse498, and avoids pitfalls and bad practices like using namespace std

- **Adheres to the C++ standard**
Member variables are initialized and undefined behavior is avoided. Also makes use of expected.

- **Appropriate Modularity**
Classes clearly serve a purpose, and there is not a case of "those who know, do" issues within member variables

- **Modern C++**
As mentioned above, expected is used and range based for loops are used when possible

- **Pointer Handling**
Makes use of good pointer memory management by using shared and unique pointers when relevant. Null pointer checks are also used to avoid de-referencing invalid objects.


Great job!

tests/InteractiveWorld/Building.cpp
finfroc3, line 1: All tests pass and look good overall, though SetCurrentLevel isn't tested (it is marked maybe_unused, though).  GetRateModifier, and SetRateModifier aren't included either, and though the setter is used in the test for ResourceProducer, the getter is never properly tested.

source/Interfaces/TrashInterface.hpp
```diff
@@ -83,25 +86,34 @@ class TrashInterface : public InterfaceBase {
...
+                    InteractiveWorldSaveManager save_manager;
+                    if (save_manager.Save(*interactive_world, "interactive_world_save.json")) {
+                        std::cout << "\nGame saved to interactive_world_save.json\n";
+                    } else {
+                        std::cout << "\nFailed to save game.\n";
+                    }
+                }
+            }
+
                 exit(0); // Quit!
```
hoangtu0701, line 117: Calling exit(0) here skips all destructors and RAII cleanup. If any other module needs to do something on shutdown it just won't happen. Setting a flag to break out of the game loop would be a lot safer.

source/Interfaces/TrashInterface.hpp
```diff
@@ -83,25 +86,34 @@ class TrashInterface : public InterfaceBase {
...
-            case 'D':
                 action_id = GetActionID("right");
                 break;
-            case 'q':
-            case 'Q':
+            case 'e':
+                action_id = GetActionID("interact");
+                break;
+            case 'q': {
+                const auto* interactive_world = dynamic_cast<const InteractiveWorld*>(&world);
```
hoangtu0701, line 106: This ties TrashInterface to InteractiveWorld specifically, which breaks the generic interface/world separation. If save-on-quit is needed, it'd be better to push that logic into the world's shutdown path or use a callback so the interface stays world-agnostic

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -40,10 +44,37 @@ class InteractiveWorld : public WorldBase {
...
+
+    bool TryInteractAdjacent(const WorldPosition& position, const AgentBase* initiator) {
+        const std::array<WorldPosition, 4> priorities = {
+                position.Right(),
+                position.Up(),
+                position.Down(),
+                position.Left(),
+        };
+
+        for (const WorldPosition& target: priorities) {
```
hoangtu0701, line 58: This scans every agent for each adjacent tile, so it's O(agents * 4) per interaction. Works fine with a small agent count but a position-to-agent map would make this way cheaper if things scale up

source/Worlds/Hub/InteractiveWorld.hpp
```diff
@@ -177,27 +217,55 @@ class InteractiveWorld : public WorldBase {
...
 
     /**
-     * Add NPC to the world
-     * @param npc NPC to add
+     * Add Town hall to the world
+     * @param th town hall
+     * @param position where to add spawn
      */
-    void AddNPC(std::shared_ptr<NPC> npc) { m_npcs.push_back(npc); }
+    void AddTownHall(TownHall& th, WorldPosition position) {
```
hoangtu0701, line 260: These four methods all do the same thing - set location and mark the tile. Could consolidate into a single method to cut the duplication.

source/Worlds/Hub/ResourceProducer.hpp
```diff
@@ -72,14 +75,14 @@ class ResourceProducer {
...
+            m_resourceSpawn->AddResource(whole);
             m_accumulator -= whole;
         }
     }
 
 private:
-    InteractiveWorldInventory& m_inventory; // World Inventory
-    std::shared_ptr<Building> m_building; // Building modifying the output rate
+    ResourceSpawn* m_resourceSpawn;
+    Building* m_building = nullptr; // Building modifying the output rate
```
hoangtu0701, line 85: Raw pointer here but the class doesn't own the building. Since the demo always passes references, storing a reference instead would make the lifetime relationship clearer.

source/Agents/AI/FetchAgent.hpp
```diff
@@ -0,0 +1,331 @@
...
+        const long long dy =
+            static_cast<long long>(a.CellY()) - static_cast<long long>(b.CellY());
+        return std::llabs(dx) + std::llabs(dy) == 1;
+    }
+
+    /// @brief BFS through walkable tiles to find the first step that moves
+    ///        `from` closer to a tile adjacent to `goal`. The goal tile itself
+    ///        is treated as reachable even though it is non-walkable, so the
+    ///        expansion terminates on the tile adjacent to the endpoint.
+    [[nodiscard]] size_t NextStepToward(const WorldGrid& grid, WorldPosition from,
```
hoangtu0701, line 262: Full BFS every tick is a lot of work for agents that shuttle the same route over and over. Caching the path until the goal or grid changes would be a nice optimization.

hoangtu0701 (commented):
Interoperability & API

Group14_main.cpp does a nice job showing the whole resource pipeline working together - spawns, producers, fetch agents, buildings, town hall, resource management menu, all in one runnable demo. The group14.md doc is also really solid, gives a clear breakdown of what each class does and how to test things manually.

A few things I noticed on the interoperability side:

1. TrashInterface calls exit(0) directly when the user hits 'q'. That skips all destructors and RAII cleanup, so any other module that needs to do something on shutdown (analytics logging, save confirmation, etc.) just never gets a chance to run. Would be cleaner to set a flag that breaks out of the game loop instead.

2. TrashInterface downcasts to InteractiveWorld inside SelectAction to handle saving. That means the interface now has a hard dependency on a specific world implementation, which kind of defeats the purpose of having a generic interface/world separation. Maybe move the save logic into the world's shutdown path or use a callback so the interface doesn't need to know what world it's in.

3. A few files still have `@note Status: PROPOSAL` but the classes are clearly functional and being used in the demo. Might want to update those.



Code Design

The class hierarchy makes sense overall. ResourceBank as a base for Building is a clean choice, the FetchAgent endpoint callback pattern is flexible, and the ResourceManagementAgent menu system is well structured. Some things that caught my eye:

1. InteractiveWorld::TryInteractAdjacent loops through every agent in the world for each of the 4 adjacent tiles. Works fine now but that's O(agents * 4) per interaction, which could get rough if the agent count grows. Even a simple position-to-agent map would bring that down to constant time per tile.

2. AddBuilding, AddResourceSpawn, AddResourceBank, and AddTownHall all do essentially the same thing - set the entity's location and mark the tile as building_id. Could probably consolidate those into one method to cut down on the repetition.

3. FetchAgent::NextStepToward runs a full BFS every single tick. For agents that are just shuttling back and forth on the same route, caching the path and only recalculating when the goal or grid actually changes would save a bunch of unnecessary work.

4. ResourceProducer holds a raw pointer to Building, but the demo always constructs producers with references. Storing a reference instead (or at least documenting the lifetime expectation) would make the non-ownership clearer and reduce the chance of a dangling pointer down the road.

5. FarmingAgent.hpp and FarmingAgent.cpp are Group 2's files. Having another group's source in your PR diff makes it a bit harder to tell what's actually Group 14's work vs. what's coming from elsewhere.

tests/InteractiveWorld/ResourceBank.cpp
finfroc3, line 1: All tests pass and everything in ResourceBank is covered.

tests/InteractiveWorld/ResourceProducer.cpp
finfroc3, line 1: All tests pass and look good, could be a good idea to make tests for ResourceSpawn as well.

tests/Agents/TestFetchAgent.cpp
finfroc3, line 1: All tests pass and test the overall functionality and intended behavior well. If there's time, I suggest including additional tests for the pathfinding (things like obstacles or having there be no valid path to origin/deposit).

tests/Agents/TestResourceManagementAgent.cpp
finfroc3, line 1: All tests pass, what is tested looks good but it is overall missing coverage for parts of ResourceManagementAgent. There exists nothing for hireable lanes, locked buildings, the functions for handling interactions, and nothing for minor functions like ClearManagedBuildings or some of the setters and getters.

demos/InteractiveWorld/Group14_main.cpp
finfroc3, line 1: Very good example driver! It runs just fine on my system. The agents, world, buildings and inventory are integrated very well, and the doc that came with was really detailed and well written. Thanks to that, I was quickly able to follow and understand the setup of everything. I do think that more comments are needed, though.

source/Agents/AI/FetchAgent.hpp
finfroc3, line 1: API is good, but more thorough documentation would be appreciated.

source/Agents/Classic/ResourceManagementAgent.hpp
finfroc3, line 1: The API is mostly well designed and self explanatory, though complex. It greatly lacks in any documentation, though. The use of raw pointers is a little sketchy as well.

source/Worlds/Hub/InteractiveWorld.hpp
finfroc3, line 1: Outside of issues touched on in the other review comments, the API is well made and easy to understand.

source/Worlds/Hub/ResourceProducer.hpp
finfroc3, line 1: Very simple, cleanly made, and well documented.

source/Worlds/Hub/Building.hpp
finfroc3, line 1: Very well documented, good job.

PR #93, opened 2026-04-21
Group 2

Cobathon:
Things in this PR:

Enemy agent behavior -
1. File: AgentFactory Skeleton Pipeline & behavior (it got messed up and partially merged to main when it was new content entirely. Anything with writing of skeleton is new)
2. Interaction in WorldBase.hpp was modified for enemy and range. This function is for the player to interact with other agents including attacking if it is an enemy
3. Weapons & Items - added modifications for the player based on the items
4. Documentation & Tests

Changes were done in Group14's branch relating to FarmingAgent class & test file.

Also review:
PathGenerator::IsPathClear (reworked)
New functions of :
PathGenerator::FindPointAway
PathGenerator::FindFurthestPoint
test cases for these are likewise at the end of PathGenerator Tests


karensuzue:
**Reviewers for Group 2:** Please select all commits starting from **[07f3bdb](https://github.com/CSE498/Spring2026-CompanyA/pull/93/commits/07f3bdb889db8ae9a846d07953622e561819874c) and earlier**

source/Agents/Classic/AgentFactory.hpp
Mehver, line 12: ### Implementation Quality

In this `AgentFactory.hpp` you used `std::abs`, so please make sure to include `<cmath>`.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/AgentFactory.hpp#L96-L100

source/Agents/Classic/AgentStats.hpp
Mehver, line 10: ### Implementation Quality

In this `AgentStats.hpp` you used `assert`, so please make sure to include `<cassert>`.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/AgentStats.hpp#L30-L36

source/Agents/Classic/AgentStats.hpp
```diff
@@ -18,10 +17,21 @@ struct AgentStats {
...
+     * Construct with baseline combat values.
+     * @param hp Max and initial HP.
+     * @param atk Base attack value.
+     * @param def Base defense value.
+     * @param range Attack range in grid units.
+     * @param level Agent level.
+     */
+    AgentStats(double hp, double atk, double def, double range, size_t level) :
+        mMaxHp(hp), mHp(hp), mAtk(atk), mDef(def), mRange(range), mLevel(level)
+    {
```
Mehver, line 33: ### Consistent Style

By the way, I might be mistaken, but just to confirm, did you run clang-format again? Because this brace placement doesn’t seem likely under the current formatting config.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/AgentStats.hpp#L30-L33

The same question arises in `Inventory.hpp`:

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L56-L61

source/Agents/Classic/PlayerFeatures/Inventory.hpp
```diff
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <array>
+#include <functional>
```
Mehver, line 20: ### Implementation Quality

In this `Inventory.hpp` you used `std::derived_from`, so please make sure to include `<concepts>`.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L56-L58

Mehver:
Hi, I’m not a reviewer for style, but I noticed that the formatting in some places seems a bit inconsistent. I’m not entirely sure, but it looks like your latest changes might not have been run through clang-format yet. There’s already a `.clang-format` configuration file in the root directory, please make sure to run it again.

source/Agents/Classic/PlayerFeatures/Inventory.hpp
```diff
@@ -87,9 +88,35 @@ class Inventory {
...
+     * Used e.g. by PlayerAgent to refresh combat stats from the hand slot.
+     */
+    void SetChangeNotifier(std::function<void()> cb) { mOnChanged = std::move(cb); }
+
+    /**
+     * Set active hotbar slot directly.
+     * @param index New hotbar index.
+     */
+    void HotBarIndexMove(int index) {
+        mCurrentHotbarSlot = static_cast<size_t>(index);
```
Mehver: ### Implementation Quality

I noticed that you pass in an `int` and then convert it using `static_cast<size_t>`. Is there a good reason for doing this? Otherwise, I think it would be more robust to change the parameter type to `size_t` directly. After all, this is a signed-to-unsigned conversion, which could lead to misuse.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L98-L105

source/Agents/Classic/PlayerFeatures/Inventory.cpp
Mehver, line 72: ### Implementation Quality

I’m not entirely sure, but it seems to me that the documented definition of `isAllOrNothing` might be the opposite of its actual behavior.

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L68-L71

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.cpp#L65-L73

source/Agents/Classic/PlayerFeatures/Inventory.hpp
Mehver: ### Implementation Quality

These two seem like they could be marked as `const`.

```c++
[[nodiscard]] bool Contains(const std::string& name) const { 
    return (mItem && mItem->GetName() == name); 
}

[[nodiscard]] bool Contains(size_t itemId) const { 
    return (mItem && mItem->GetId() == itemId); 
}
```

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L186-L187

Mehver: ### Implementation Quality

This return value is a raw pointer; you might consider adding a `const` qualifier. Please decide based on the intended semantics.

```c++
const Item* GetHand() const
```

https://github.com/CSE498/Spring2026-CompanyA/blob/5ac4d65d4f3e8159edc07788f8ba65fee9d2ea47/source/Agents/Classic/PlayerFeatures/Inventory.hpp#L134

source/Agents/Classic/AgentFactory.hpp
Mehver, line 135: ### Testing and Error Checking

I didn’t see any related tests for this function. If it indeed doesn’t have any, I think it would be worth adding them.

https://github.com/CSE498/Spring2026-CompanyA/blob/879f8a3d5c02b8432df8ad77609f743b762c86a1/source/Agents/Classic/AgentFactory.hpp#L134

Mehver, line 137: ### Testing and Error Checking

This function is only testing in `source/Worlds/DemoG2/DemoSimpleWorldG2.cpp` (Line 326). If this is a demo that will continue to be maintained, that might already be sufficient, so this is just a reminder for you to double-check.

https://github.com/CSE498/Spring2026-CompanyA/blob/879f8a3d5c02b8432df8ad77609f743b762c86a1/source/Agents/Classic/AgentFactory.hpp#L136-L137

https://github.com/CSE498/Spring2026-CompanyA/blob/484535be012d35c77afef4fe4490fa026e8428d3/source/Worlds/DemoG2/DemoSimpleWorldG2.cpp#L320-L327

Also, there is now a `/demos/` directory in the root. I suggest placing all demos in this folder to make maintenance easier.

source/Agents/Classic/PlayerFeatures/Inventory.hpp
Mehver: ### Testing and Error Checking

As I mentioned in https://github.com/CSE498/Spring2026-CompanyA/pull/93#discussion_r3126697163, you might also consider adding other types of invalid inputs in your tests (`tests/Agents/TestInventory.cpp`), such as floating-point numbers or negative values, so that this implementation can be ruled out during the TDD phase.

https://github.com/CSE498/Spring2026-CompanyA/blob/879f8a3d5c02b8432df8ad77609f743b762c86a1/tests/Agents/TestInventory.cpp#L334-L335

source/Agents/Classic/PlayerFeatures/Inventory.cpp
Mehver, line 12: ### Testing and Error Checking

https://github.com/CSE498/Spring2026-CompanyA/blob/879f8a3d5c02b8432df8ad77609f743b762c86a1/source/Agents/Classic/PlayerFeatures/Inventory.cpp#L12-L21

This part includes a check for pointers, which is good. However, this seems to be a core and frequently used interface, so it would be better to expand the tests in `TestInventory.cpp` as well. For example, by passing `nullptr` to further define its boundaries and maintain its robustness.

source/Agents/Classic/AgentFactory.cpp
gudimaya, line 36: So I'm not fully sure, but if assert fails then it could lead to crashing the program. Small suggestion could be:

if (!grid.IsWalkable(nextPos)) {
    return Failure;
}

This way it wont crash if anything goes south, but an additional check could be added to try and run the process again.

gudimaya: This is great and all, but maybe we can add test case where the player might be equal to null, in case of bugs or mess ups. 

auto player = world.GetPlayer();
if (player == nullptr || !player->IsAlive()) {
    return Failure;
}

gudimaya, line 90: The code comments say that an invalid movement direction may return an empty optional, but the code should explicitly test for that before requesting an action ID.

auto actionName = MovementTypes::GetActionName(dir);
if (!actionName) {
    return Failure;
}

size_t actionId = enemy.GetActionID(*actionName);
if (actionId == 0) {
    return Failure;
}

gudimaya, line 3: I think in general this is really good and is pretty much release ready. The only fixes I really recommended are like safety fixes. You could add some testing as well such as adding tests where the player is dead to verify that enemies do not attack, chase, or select movement actions.

IsPlayerInRange -> Failure
AttackPlayer -> Failure
ChasePlayer -> Failure

Could also add a blocked map test where there is no path between the enemy and player. The chase node should fail cleanly and should not try to phase through a wall or get stuck etc. 

ChasePlayer -> Failure
selected_action is not set

source/Agents/Classic/PlayerAgent.cpp
gudimaya: The API assumes that "stay" is always a valid registered action. If the action table changes, an invalid input could produce actionID = 0. It may be better to expose a named invalid action value or check for failure explicitly.

size_t action = GetActionID("stay");
if (action == 0) {
    // handle invalid action
}
return action;

source/core/item/ItemConsumableDefense.hpp
gudimaya: Should this be made const?

int GetDefense() const { return m_defense; }

source/core/item/ItemWeapon.hpp
```diff
@@ -18,18 +18,25 @@ class ItemWeapon : public Item {
     double m_hit_bonus; // Bonus the weapon has to hit
 
 public:
+    /**
+     * Base weapon item constructor.
+     * Initializes weapon modifiers to zero so derived weapons can opt in.
+     */
     ItemWeapon(size_t id, const std::string& name, const std::string& image_path, int gold, const WorldBase& world) :
-        Item(id, name, image_path, gold, world) {}
+        Item(id, name, image_path, gold, world), m_range(0.0), m_damage(0.0), m_hit_bonus(0.0) {}
```
gudimaya, line 26: The constructor seems to contain two conflicting initializer lists. This should probably be cleaned up because it could create an unclear and invalid API for constructing weapon items.

ItemWeapon(size_t id, const std::string& name, const std::string& image_path, int gold, const WorldBase& world)
    : Item(id, name, image_path, gold, world),
      m_range(0.0),
      m_damage(0.0),
      m_hit_bonus(0.0) {}

gudimaya (commented):
Interoperability & API and Testing & Error Checking. 

Overall this is really great, I left a few possible fixes and suggestions that honestly not all need to be implemented, but they could clean up the final product for sure. let me know of any questions!

source/Agents/Classic/AgentFactory.hpp
Cobathon: Good catch, added <cmath> in AgentFactory.hpp

source/Agents/Classic/AgentStats.hpp
Cobathon: Thanks, added <cassert> in AgentStats.hpp for assert.

source/Agents/Classic/PlayerFeatures/Inventory.hpp
Cobathon: Added <concepts> for std::derived_from in Inventory.hpp.

source/Agents/Classic/PlayerFeatures/Inventory.hpp
Cobathon: I changed HotBarIndexMove to size_t and made the setter robust with hotbar wrap

source/Agents/Classic/PlayerFeatures/Inventory.hpp
Cobathon: Both Contains overloads are now const.

source/Agents/Classic/AgentFactory.cpp
Cobathon: I replaced the answer-style assumptions in chase flow with failure checks, so no crashing happens.

Cobathon: Thank you for the suggestion! Adjustments are in the works

Cobathon: Added an explicit invalid-action-name guard before action lookup (empty -> Failure) and preserved the action_id == 0 check

source/Agents/Classic/PlayerAgent.cpp
Cobathon: I made the default stay-action lookup explicit in code so failure handling is clearer

source/core/item/ItemConsumableDefense.hpp
Cobathon: Yes, change has been made

source/core/item/ItemWeapon.hpp
Cobathon: Thank you for the suggestion

kianamay (commented):
Overall, the modules are clearly meant to work together, and the code does a good job showing interaction between agents, world logic, items, and pathfinding. The biggest interoperability concern is that many contracts are implicit instead of documented. For example, AgentFactory.cpp assumes world.GetPlayer() is valid, PlayerAgent.cpp assumes inventory hand items can be checked through the item inheritance hierarchy, and several files rely on string-based action names or numeric action IDs. These work internally, but other groups may misuse them without clearer API documentation.

Informative identifiers:
Overall, identifiers are mostly clear. Names like AgentFactory, CreateEnemyGoblin, RefreshCombatFromHand, BuildDiamondNeighborsByRange, and FindShortestPath communicate purpose well. The main issue is that some cross-module values are still unclear, especially raw action IDs, string action names, and priority values in WorldBase.hpp.

Avoids magic numbers:
This area could be improved. Some constants are named well, such as SKELETON_MAX_STEP_AWAY_COUNT, MAX_SEARCH_DISTANCE, EP, and STEP_SIZE. However, there are still places where values like action IDs, interaction priorities, range values, and default item stats could be more clearly represented with named constants or enums.

Intuitive and functional interface:
The interfaces are generally usable and modular. AgentFactory makes enemy creation easier, Inventory exposes useful item operations, and PathGenerator separates pathfinding from AI behavior. The biggest issue is that some interfaces are easy to misuse because they rely on raw pointers, strings, or primitive return values. For example, GetHand() returning Item*, action lookup through strings, and GetPlayer() assumptions should be better documented or made safer.

Informative comments:
Overall good, however some comments could be made more detailed  for API-level documentation, especially in files meant to be used by other groups, such as AgentFactory.hpp, WorldBase.hpp, and Inventory.hpp.

Consistent style:
Style is mostly consistent, but there are some formatting inconsistencies in brace placement and spacing across files like AgentStats.hpp, PathGenerator.cpp, and tests. Running clang-format would help make the code easier for other groups to scan and review.

Visual appearance:
The code is generally readable and separated into logical sections. Larger files like AgentFactory.cpp and PathGenerator.cpp are harder to scan because they contain dense logic and multiple responsibilities. Breaking some behavior-tree or pathfinding helper logic into smaller named functions could improve readability and make the module easier for other groups to use.

ncharters:
### Implementation Quality:
#### Compilation:
I received an error during the build process that stopped your code from building. I think it's a new change since I didn't notice the specific line in the diffs for the commits that you asked for us to look at, but still worth noting. In the AttackPlayer function in AgentFactory, this line of code - `player->TakeDamage(dmg);` - gives the following error: 

> error: passing 'const cse498::AgentBase' as 'this' argument discards qualifiers


#### Namespaces:
All code is properly contained within the cse498 namespace.

#### C++ standard:
I didn't notice any instances where there may be undefined behavior in your module, but as mentioned above, the code does fail to build currently.

#### Modularity:
The modularity of the code is well done, all of the agent and item classes derive from base classes and override any functions that they need to alter the functionality of. Additionally, there are plenty of good helper functions used throughout that eliminate redundant code.

#### Pointers:
Good use of smart pointers throughout where applicable.

#### Const:
I noticed a lot of functions and variables have been updated to reflect proper const properties, so good work on that as well.


### Advanced C++:
Great job incorporating the advanced topics into the module, I noticed multiple good usage of templated functions, as well as std::move, and any values known at compile time were properly marked as constexpr.

source/Agents/Classic/PlayerFeatures/Inventory.cpp
ZodiacalBee: added

ZodiacalBee: I flipped the sign to fix it a while ago... It seems right..??
if true && the amount isn't sufficient to remove then return amount otherwise remove all and 0 will be returned?

source/Agents/Classic/AgentFactory.cpp
ZodiacalBee: I'm confused on this? what? When could it ever return empty?

ZodiacalBee: testing for that stuff is done within the general test cases or should have been. I didn't write them, but I can tell you that most of that is tested from skeleton movement and attacking which would fail if any one part of those broke

ZodiacalBee: no phasing through the wall should ever occur. selected_action is reset every turn so I don't need to do anything in here. Shortest path obeys the given map and ensures there is a valid path and hitbox for attacking checks all times from center point of tile to other center point of time to see if obstacles are in the way

source/Agents/Classic/AgentFactory.hpp
ZodiacalBee: I'm not sure what this agent was ever used for and it doesn't test anything I would want to see. Hence, as I didn't write or see a use for it I didn't want to write meaningful tests as I think it just goes back and forth (2 tiles) not actually patrolling a path. We should probably delete it or make it better. I never decided and never did either.

ZodiacalBee: Yeah I should've rewrote the demo. I just didn't have time to rewrite it all because it could've been much shorter and this could've been deleted. but I was too busy with another project. It feels pointless to rewrite the demo at this point to fix everything. I'd rather focus on fixes for the final project

source/Agents/Classic/AgentStats.hpp
ZodiacalBee: interestingly enough the comment blocks it in inventory.hpp
but I changed it anyway

PR #99, opened 2026-04-29
Fixing Group 2 Test Cases

ZodiacalBee:
3 test cases fail left in replay driver..

I'm not sure what caused some of these test cases to literally be written wrong and become bad ... really strange. I wish I knew what happened.


source/Agents/Classic/PlayerFeatures/Inventory.hpp
```diff
@@ -205,7 +205,7 @@ class Inventory {
             return Insert(quantity);
         }
         size_t Insert(size_t quantity) {
-            if (mItem && mItem->IsUnique())
+            if (!mItem || mItem->IsUnique())
```
ZhixiangMiao, line 208: I like this code design, since your previous code used an and to check if mItem is valid and mItem can point to IsUnique(), but it was too vague and may repeat like double-checking the same mItem. But in your updated code, you check if either mItem is invalid or otherwise it will point to IsUnique() to get rid of the risk of repeat.

karensuzue (approved):

source/Agents/Classic/FarmingAgent.cpp
```diff
@@ -12,231 +12,193 @@
...
+    double bestDistance = std::numeric_limits<double>::infinity();
 
-        for (const WorldPosition& candidate : candidates)
-        {
-            // only consider walkable tiles
-            if (!grid.IsWalkable(candidate))
-            {
-                continue;
-            }
+    for (const WorldPosition& candidate: candidates) {
```
ZhixiangMiao, line 51: Code Design: Yes, this filters blocked or non-walkable adjacent tiles. However, it still chooses the best candidate based on the Manhattan distance rather than actual path reachability. It would be safer to choose the adjacent work tile based on whether a valid path exists, or to compare candidates by shortest path length rather than by Manhattan distance.

source/Agents/Classic/FarmingAgent.cpp
```diff
@@ -12,231 +12,193 @@
...
-        if (path.Size() < 2)
-        {
-            return 0;
-        }
+    const WorldPosition currentPosition = GetPosition();
+    const WorldPosition nextPosition = path.At(1);
 
-        const WorldPosition currentPosition = GetPosition();
-        const WorldPosition nextPosition = path.At(1);
+    // Convert tile in path to movement action
```
ZhixiangMiao, line 77: Code Design: Try not to use hard-coded strings such as"up", use things from WorldActions class instead. Since hard-coding could cause bugs in the future when anything changes.

tests/Agents/TestSkeleton.cpp
ZhixiangMiao, line 1: Advanced C++: Try to add more comments to describe your tests. This would help the reader to understand better.

ZhixiangMiao:
Advanced C++: Overall, better add more comments to describe more details about your code. This can help readers to understand better, and can also help you to organize your code in the future.

ZhixiangMiao (commented):
Overall, the code looks great, however, there are still a few problems. 

**Code Design:**
When you try to find an adjacent work tile, I suggest choosing the best candidate based on the actual path reachability. You should try each walkable adjacent tile until a valid path exists. And also try not to hard-code (like you did just input the hard-coded string "up"), you should find and see if there are any usable methods that can do similar things. 

**Advanced C++:**
I suggest you add more comments to your code so readers can better understand what you are trying to do. Adding comments is also a good practice for you when you organize/debug your code in the future, since it can be easier for you to remember what was going on based on comments. 

Again, the code barely used modern C++ methods, and I suggest using more to make your code cleaner and more efficient. Modern C++ methods, such as lambda, are a great way to make and implement a simple function, and it is more understandable and useful.

ZhixiangMiao (commented):
**Advanced C++:** I didn't see any lambda usage, and template usage was too little, and you should use template typename T for more generic use. However, I saw you used some modern C++ methods, but they are barely used in one file. I suggest using more modern C++ methods by finding some spot to add them into your code or replacing some redundant, complex code with modern C++ methods, since they could be cleaner and more efficient.

